PR #2, opened 2016-09-28
Update gradlew

oroszgy:
Fixed the gradlew file to make it actually work.

### DIFF
--- a/gradlew
+++ b/gradlew
@@ -84,6 +84,10 @@ Please set the JAVA_HOME variable in your environment to match the
 location of your Java installation."
 fi
 
+# Create the neccessary templates for IPython
+mkdir ~/.ipython/kernels/kotlin
+touch ~/.ipython/kernels/kotlin/kernel.json
+
 # Increase the maximum file descriptors if we can.
 if [ "$cygwin" = "false" -a "$darwin" = "false" ] ; then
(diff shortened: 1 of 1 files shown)


gradlew
```diff
@@ -84,6 +84,10 @@ Please set the JAVA_HOME variable in your environment to match the
 location of your Java installation."
 fi
 
+# Create the neccessary templates for IPython
+mkdir ~/.ipython/kernels/kotlin
```
happy15: "mkdir -p" would be better cause in my case I don't even have the kernels folder.

gradlew
```diff
@@ -84,6 +84,10 @@ Please set the JAVA_HOME variable in your environment to match the
...
+touch ~/.ipython/kernels/kotlin/kernel.json
```
happy15: And there is no need to touch it since the build.gradle script will create it upon writing.

hastebrot:
I had to manually create the `~/.ipython/kernels/kotlin` directory under Ubuntu 16.04. Besides this, the Kotlin kernel works good so far, even if Linux is not officially supported right now.

This fix should definitely go into `build.gradle` instead of `gradlew`. Gradle's `wrapper` task could always overwrite this change.

ligee:
The directory creation is properly addressed by https://github.com/ligee/kotlin-jupyter/pull/9 (merged)

PR #33, opened 2019-12-12
Add TeamCity integration

ileasile:


distribution/conda/meta.yaml
```diff
@@ -0,0 +1,31 @@
+{% set data = load_setup_py_data() %}
+
+package:
+  name: kotlin-jupyter-kernel
```
ileasile: Cannot get name from `setup.py` because of conda bug: https://github.com/conda/conda-build/issues/1379

build.gradle
```diff
@@ -1,6 +1,8 @@
-import java.nio.file.Files
```
nikitinas: [build.gradle](build.gradle) becomes too heavy and needs to be reogranized. We may extract some part of it into `buildSrc` folder as described [here](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources)

ileasile: Moved part of this file to `distrib.gradle`

build.gradle
```diff
@@ -85,74 +147,330 @@ jar.manifest.attributes(
...
+class CondaTaskSpec extends TaskSpec {
+    String username
+    CondaCredentials credentials
 }
 
-createTaskForSpecs(true)
-createTaskForSpecs(false)
+ext.condaCredentials = new CondaCredentials(
+        username: condaUserStable,
+        password: condaPasswordStable
```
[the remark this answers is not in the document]
ileasile: Done

Issue #31, opened 2019-12-07
Config editor / network based configuration


altavir:
Compile-time configuration repository seems to be rather limited. We probably will want to add configuration locally (Jupyter lab has special configuration pages for extensions) and in future, there should be central repository for packages ready for notebook so people could update them without installing new version. 

Low priority.

ileasile:
@altavir  You still can use `@file:DependsOn()` and `@file:Repository()` in runtime. Why doesn't it suit you?

altavir:
Because configuration allows to define renderers and possibly other things, which is quite important. Also what average notebook user will want is a number of community-chosen and approved modules with centralized description. Power user can attach his own dependencies via `DependsOn`, but more general public will prefer to use something that is tested. I think that Kotlin notebook now has significant advantage over regular Python ones because we can define dependencies and versions inside the notebook and make it stable without bothering with external environment.

PR #34, opened 2019-12-17
Library configuration improvements

nikitinas:
1. Split library descriptors into separate jsons
2. Load local library descriptors from "<user>/.jupyter_kotlin/libraries" directory
3. Download library descriptors from master branch of "https://github.com/Kotlin/kotlin-jupyter" repository
4. Cache downloaded descriptors in "<user>/.jupyter_kotlin/cache/libraries" directory
5. Make descriptors loading async
6. Check library descriptor format version and suggest updating kernel if format was changed

Fixes https://github.com/Kotlin/kotlin-jupyter/issues/31


src/main/kotlin/org/jetbrains/kotlin/jupyter/config.kt
```diff
@@ -1,9 +1,23 @@
...
 import kotlin.script.experimental.dependencies.RepositoryCoordinates
 
+val LibrariesDir = "libraries"
+
+val LocalSettingsPath = Paths.get(System.getProperty("user.home"), ".jupyter_kotlin").toString()
+
+val GitHubApiHost = "api.github.com"
+val GitHubRepoOwner = "kotlin"
+val GitHubRepoName = "kotlin-jupyter"
+val GitHubApiPrefix = "https://$GitHubApiHost/repos/$GitHubRepoOwner/$GitHubRepoName/contents/"
```
ileasile: What about adding branch here? It would be helpful when changing this code and testing new functionality. I'd even made it a system property.

src/main/kotlin/org/jetbrains/kotlin/jupyter/config.kt
```diff
@@ -27,16 +53,99 @@ class LibraryDefinition(val dependencies: List<String>,
...
+    val librariesMap = readLibraries(LocalSettingsPath).toMap().orEmpty().toMutableMap()
+
+    val address = GitHubApiPrefix + LibrariesDir
+    val response = catchAll { khttp.get(address) }
+    if (response != null && response.statusCode == 200) {
+        response.jsonArray.forEach {
+            val o = it as JSONObject
+            val filename = o["name"] as String
+            if (filename.endsWith(".json")) {
+                val libName = filename.substring(0, filename.length - 5)
```
ileasile: What about using regex here to avoid magic `5`?

src/main/kotlin/org/jetbrains/kotlin/jupyter/repl.kt
```diff
@@ -88,9 +88,10 @@ class ReplForJupyter(val scriptClasspath: List<File> = emptyList(),
             val kt = KotlinType(receiver.javaClass.canonicalName)
             implicitReceivers.invoke(listOf(kt))
 
-            val classes = listOf(/*receiver.javaClass,*/ ScriptTemplateWithDisplayHelpers::class.java)
-            val classPath = classes.asSequence().map { it.protectionDomain.codeSource.location.path }.joinToString(":")
-            compilerOptions.invoke(listOf("-classpath", classPath, "-jvm-target", "1.8"))
+            //val classes = listOf(/*receiver.javaClass,*/ ScriptTemplateWithDisplayHelpers::class.java)
+            //val classPath = classes.asSequence().map { it.protectionDomain.codeSource.location.path }.joinToString(":")
```
ileasile: Are these comments still needed? Where did this logic go?

src/main/kotlin/org/jetbrains/kotlin/jupyter/config.kt
```diff
@@ -27,16 +53,99 @@ class LibraryDefinition(val dependencies: List<String>,
...
+                if (!librariesMap.containsKey(libName)) {
```
ileasile: Want to encapsulate this check somehow, because the same check is done below... But it's optional.

src/main/kotlin/org/jetbrains/kotlin/jupyter/config.kt
ileasile: What about using regex here to avoid magic `5`?

src/main/kotlin/org/jetbrains/kotlin/jupyter/config.kt
ileasile: What about adding branch here? It would be helpful when changing this code and testing new functionality. I'd even made it a system property.

src/main/kotlin/org/jetbrains/kotlin/jupyter/repl.kt
nikitinas: Additional classpath is not needed, because shared classes for completion have already been moved to jupyter-lib module that is isolated from kernel internals. -Jvm-target needs to be restored, thanks for pointing this out

nikitinas: -jvm-target is already applied below this code. So, comment can be removed

src/main/kotlin/org/jetbrains/kotlin/jupyter/config.kt
nikitinas: done, thanks

src/main/kotlin/org/jetbrains/kotlin/jupyter/config.kt
nikitinas: Added

src/main/kotlin/org/jetbrains/kotlin/jupyter/config.kt
nikitinas: done

src/main/kotlin/org/jetbrains/kotlin/jupyter/repl.kt
```diff
@@ -167,7 +171,7 @@ class ReplForJupyter(val scriptClasspath: List<File> = emptyList(),
 
     init {
         // TODO: to be removed after investigation of https://github.com/kotlin/kotlin-jupyter/issues/24
-        eval("1")
+        doEval("1")
```
ileasile: Is it just an optimization, or `eval()` will no longer work here?

nikitinas: `eval()` uses library descriptors, that are now computed asynchronously, so we don't want to await them here. They will be awaited on first real execute request.

src/main/kotlin/org/jetbrains/kotlin/jupyter/config.kt
```diff
@@ -86,40 +82,119 @@ fun parseLibraryName(str: String): Pair<String, List<Variable>> {
...
-                if (!librariesMap.containsKey(libName)) {
-                    val url = o["download_url"].toString()
-                    val res = catchAll { khttp.get(url) }
-                    if (res != null && res.statusCode == 200) {
-                        log.info("Loading '$libName' descriptor from '$url'")
-                        librariesMap[libName] = res.jsonObject.toJsonObject()
-                    }
-                }
-            }
+            if ("""[\w-]+.$LibraryDescriptorExt""".toRegex().matches(filename)) {
```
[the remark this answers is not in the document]
nikitinas: I forbid spaces to have cleaner syntax in "%use lib" command. Allowed dots, fixed bug with '\.', thanks. Regex moved out of lambda

src/main/kotlin/org/jetbrains/kotlin/jupyter/util.kt
```diff
@@ -22,11 +27,24 @@ fun <T> Logger.catchAll(msg: String = "", body: () -> T): T? = try {
...
+fun <T, R> Deferred<T>.asyncLet(selector: suspend (T) -> R): Deferred<R> = this.let {
+    GlobalScope.async {
+        selector(it.await())
+    }
+}
+
+fun <T> Deferred<T>.awaitBlocking(): T = if (isCompleted) getCompleted() else runBlocking { await() }
+
+fun String.parseIniConfig() =
+        split("\n").map { it.split('=') }.filter { it.count() == 2 }.map { it[0] to it[1] }.toMap()
```
ileasile: Why isn't some INI parser used? In this implementation comments (`;`) are not handled properly. I understand that `.properties` has only one line, and it may be left as is this time.

PR #37, opened 2019-12-25
Use gradle variables in kernel source

ileasile:
- Added possibility of using properties from Gradle in source code
- Fixed implementation version


src/main/kotlin/org/jetbrains/kotlin/jupyter/repl.kt
```diff
@@ -38,7 +38,8 @@ class ReplCompilerException(val errorResult: ReplCompileResult.Error) : ReplExce
 }
 
 class ReplForJupyter(val scriptClasspath: List<File> = emptyList(),
-                     val config: ResolverConfig? = null) {
+                     val config: ResolverConfig? = null,
+                     val properties: RuntimeKernelProperties = RuntimeKernelProperties()) {
```
nikitinas: Why does ReplForJupyter need RuntimeKernelProperties? They are not used within it

PR #60, opened 2020-04-29
kotlinVersion to 1.3.71 to make teamCity build great again (or not)

fbrunacci:
ext.kotlinVersion version correction 
1.3.70-eap-3 not longer available


ileasile:
We will move to dev version of Kotlin 1.4 in #46. It is needed because we will use some new, recently developed, features of scripting REPL. Teamcity build is broken because of another reason)

PR #65, opened 2020-05-01
Fix syntax error in samples

breandan:


ileasile:
I've reset execution counters in this notebook to make diff smaller

PR #89, opened 2020-07-22
Refactor libraries handling, add handling of descriptors from different sources

### LINKED ISSUE #70: Add optional version tag to use directive
Currently, `use` uses current version of the module descriptor from the distribution or the web. As it is, it solve significant portion of the problem with notebook reproducability, but sooner or later we will move to newer versions of artifacts mentioned in `use` and old notebooks could stop working.

I propose to add a version/variant qualifier to the module name like `kmath@0.1.3` (any other symbol could be used instead of `@`. Different version of the module descriptor could be stored in the same file and if no qualifier is provided, resolver should default empty one. Also it is possible to introduce nested qualifiers in case we will need something more complicated: `kmath@dev@0.1.4-dev-2` (default dev version is provided for `kmath@dev`). Nested descriptors are quite easy to do with json.

ileasile:
Also, get rid of several singletons in config and replaced
them with factory and free functions.

Fixes #70


README.md
```diff
@@ -128,6 +129,25 @@ Several libraries can be included in single `%use` statement, separated by `,`:
...
+%use krangl@dev(0.10)
+```
+Other options are resolving library descriptor from a local file or from remote URL:
+```
+// Load library from file
+%use mylib@file[/home/user/lib.json]
+// Load library descriptor from a remote URL
+%use herlib@url[https://site.com/lib.json]
+// You may omit library name for file and URL resolution:
+%use @file[lib.json]
```
altavir, line 149: In this case, the path is relative. Is it relative to the current notebook scope or to notebook parent directory? I think that the later is preferred since one can ship descriptors alongside the notebook.

ileasile: It's relative to current working directory, which is the directory from where you start your jupyter client

altavir: It would make sense if it would be relative to the notebook file. This way it won't matter where you start the notebook.

ileasile: As I can figure out from [this SO answer](https://stackoverflow.com/a/52119628/5488904) it's impossible to obtain the notebook path without extending the messaging protocol :( Regarding your offer about resolution against parent directory, I think it's a bit unnatural: can't really figure out the case. Anyway, you may go to the parent directory via `..`, and it will work

altavir: OK then. My idea was to be able to put descriptor in the same directory as the notebook itself and do not think about where do you start notebook from, but it is not so important to waste effort.

PR #92, opened 2020-08-05
Adds support for kotlin-spark-api

asm0dey:


libraries/kotlin-spark.json
```diff
@@ -0,0 +1,35 @@
...
+    "org.apache.spark:spark-repl_2.12:$spark",
+    "org.apache.spark:spark-sql_2.12:$spark",
+    "org.scala-lang:scala-library:$scala",
+    "org.scala-lang:scala-reflect:$scala",
+    "org.scala-lang:scala-compiler:$scala",
+    "org.scala-lang.modules:scala-xml_2.11:1.2.0",
+    "commons-io:commons-io:2.5"
+  ],
+  "imports": [
+    "import org.jetbrains.spark.api.*",
```
ileasile, line 17: Remove `import` from here

libraries/kotlin-spark.json
```diff
@@ -0,0 +1,35 @@
+{
+  "properties": {
+    "scala": "2.12.10",
+    "spark": "3.0.0",
+    "kotin-spark": "0.3.0"
+  },
+  "dependencies": [
+    "org.apache.spark:spark-repl_2.12:$spark",
```
ileasile, line 8: It seems to me that scala version (2.10, 2.11, 2.12) should be a property, and the same version should be used for all dependencies

asm0dey: Currently, only 2.12 is supported, so I think we should leave it as is at the time being.

Issue #88, opened 2020-07-19
Cell and notebook reflection API


altavir:
This issue is actually an extension of #78. The idea is to have interfaces that define cell and notebook status. Those interfaces should be placed in an external artifact that is available to plugin and library developers. When inside the notebook the instances of those interfaces could be obtained as implicit calls. My idea is that the whole code executed in a cell should be done with a cell receiver. During notebook execution, this cell receiver is substituted by the actual cell object.

The following features should be supported.

**Cell:**
- [ ] `val id: String`. A unique cell id.
- [ ] `val notebook: Notebook`. Get a reference to a current notebook.
- [ ] `val/var text: String`. The code inside the cell (see #76 ).
- [ ] `val output: MIMEContainer`. The output as MIMEContainer.
- [ ] `val rawOutput: String`. Raw output.
- [ ] Specific notebooks controls like hide input/output.
- [ ] Execution controls. Maybe asynchronous cell result.

**Notebook:**

- [ ] Kernel handle.
- [ ] A collection of resource handles with the possibility to inject new resources in runtime (see #74 ). The resources must provide asynchronous load handles or a javascript promise name so one could organize asynchronous loading of widgets. 
- [ ] Find a `Cell` by id.

breandan:
May be unrelated, but it would be nice to have hooks into the Jupyter notebook rendering system in order to display custom result types (e.g. [`Javascript`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.Javascript), [`Latex`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.Latex), [`Markdown`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.Markdown), [`SVG`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.SVG)) similar to how the Python API is implemented. Skimmed the code and it looks like [some](https://github.com/Kotlin/kotlin-jupyter/blob/a742c83884dea06f4c5ed65440c8a90128fd3980/jupyter-lib/src/main/kotlin/jupyter/kotlin/results.kt) are already supported, but perhaps I missed the other MIME types.

Issue #78, opened 2020-05-31
Publish the cell and display layout API as a separate artifact.


altavir:
I mean things like `TypedMIMEContainer` etc. It is required to be able to make plugins. I also think that there should be an abstraction for the current JSON module specification. So things like init code could be loaded directly from the jar.

ileasile:
This thing was implemented, see https://github.com/Kotlin/kotlin-jupyter/blob/master/docs/libraries.md for documentation

Issue #76, opened 2020-05-29
Support for input/output caching


breandan:
Similar to the [Python kernel](https://ipython.readthedocs.io/en/stable/interactive/reference.html#input-caching-system), it would be nice to support input/output caching. Is it possible to access the contents of other cells programmatically?

ileasile:
Hello @breandan! Yes, you can. `res1`, `res2` etc. will do the thing

ileasile:
Ah... Inputs caching is not supported. Only results are saved.

altavir:
@ileasile I failed to access results by `resX`. Is there a documentation for that? Also I agree with @breandan here, it will be nice to have. As I mentioned in another issue, beakerX has this feature and it is rather powerful. What we would like to have is the "reflective" access to both cell content and cell output via cell unique identifier. In the end we would like to have full control like [executing other cells](https://github.com/twosigma/beakerx/blob/fd8042db59f7d29dc16171592ed434f49da2dc27/doc/groovy/RunCells.ipynb) and external API artifact for that so we could easily write new notebook plugins with custom functionality.

Issue #20, opened 2018-10-01
Allow for multiple cell outputs


holgerbrandl:
Similar to python in would be nice if multiple outputs per cell would be supported.

See https://stackoverflow.com/questions/34398054/ipython-notebook-cell-multiple-outputs

altavir:
BeakerX has a feature named [OutputContainer](https://github.com/twosigma/beakerx/blob/69cf4d47d623dcb10087d6962808d020b9bf01f9/doc/groovy/OutputContainers.ipynb) it allows to define a custom layout for output cell and put multiple outputs in there. It is extremely useful, especially backed by appropriate API. I think it is quite easy to do by defining MIME type for those outputs (interactivity could be added later).

My proposal is the following (I actually designed and used something very similar for DataForge). Add implicit object called, say, `val cell: CellOuptut` for each cell. `CellOuptut` has a number of named outputs inside it including default output and a `Layout`, which defines how different outputs are rendered together. Cell result is automatically placed in the default output and if there is only one output, it is rendered as is without a container. The user can explicitly call something like `cell.render(target = "output name", obj = myObject, renderer = <renderer and renderer options>)` and the object is put into specific output. Also user can use `cell.layout{}` for layout configuration.

The `CellOutput` should not propagate between cells (each cell has one of its own), but it would be also useful to have global `CellManager` object, that would allow dynamic access to any cell output (of-course we will need to name cells or get references to them like `val id = cell.id` or `cell.name = "ddd"`). In that case and output could be located by two names: cell name/id and output name. It is done like this in DataForge and is a very flexible solution.

altavir:
Another thing solved by output control is a rendering from inside of the cell. In current implementation, the only way to make a plot is to return it from the cell. Obviously, sometimes we want to call it from inside the cell like `plot.show()`, and return something else, or show multiple plots in the same output.

ileasile:
So, there is `DISPLAY()` method which may be called multiple times per cell. You may also update displays with `UPDATE_DISPLAY()`. To list all notebook/cell displays, use `org.jetbrains.kotlinx.jupyter.api.Notebook#getAllDisplays` and `org.jetbrains.kotlinx.jupyter.api.CodeCell#getDisplays` correspondingly.

PR #99, opened 2020-09-03
Notebook API

ileasile:
This PR adds notebook API which is published as a separate artifact and may be used with any Kotlin library to integrate it with this kernel.
Some highlights:
- `notebook` property is now availible in every cell, and it exposes the API. You may define, say, `fun f(nb: Notebook<*>)` and then call it in the notebook like `f(notebook)`
- `Notebook` interface gives an access to cells and execution results
- `CodeCell` gives an access to code, streams and results of a given cell
- `Renderable` interface, once implemented by your custom class, makes it renderable
- `DisplayResult` interface is a "final" representation of your objects to be displayed with `display_data`
- displayed data may be also updated by `notebook.host.updateDisplay` function

Relates to #20, #76, #88, #78


jupyter-lib/kotlin-jupyter-api/src/main/kotlin/org/jetbrains/kotlin/jupyter/api/api.kt
```diff
@@ -0,0 +1,74 @@
...
+     */
+    val streamOutput: String
+
+    /**
+     * Previously evaluated cell
+     */
+    val prevCell: CodeCell?
+}
+
+interface KotlinKernelHost {
```
altavir, line 43: Javadoc is missing for this one.

jupyter-lib/kotlin-jupyter-api/src/main/kotlin/org/jetbrains/kotlin/jupyter/api/api.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    fun scheduleExecution(code: Code)
+
+    fun execute(code: Code): Any?
+
+    fun executeInit(codes: List<Code>)
+}
+
+interface ResultsAccessor {
+    operator fun get(i: Int): Any?
```
altavir, line 56: As far as I can understand this hold asynchronous results. I think it is worth using coroutines here.

jupyter-lib/kotlin-jupyter-api/src/main/kotlin/org/jetbrains/kotlin/jupyter/api/api.kt
```diff
@@ -0,0 +1,74 @@
...
+}
+
+interface RuntimeUtils: JavaVersionHelper
+
+interface Notebook<CellT: CodeCell> {
+    val cells: Map<Int, CellT>
+    val results: ResultsAccessor
+    val host: KotlinKernelHost
+
+    fun prevCell(before: Int): CellT?
```
altavir, line 66: I think it should be renamed to `history` or something like that

jupyter-lib/kotlin-jupyter-api/src/main/kotlin/org/jetbrains/kotlin/jupyter/api/api.kt
```diff
@@ -0,0 +1,74 @@
...
+    val thisCell: CellT?
```
altavir, line 67: currentCell?

jupyter-lib/kotlin-jupyter-api/src/main/kotlin/org/jetbrains/kotlin/jupyter/api/libraryConnection.kt
```diff
@@ -0,0 +1,47 @@
+package org.jetbrains.kotlin.jupyter.api
+
+typealias TypeName = String
+typealias Code = String
+
+data class TypeHandler(val className: TypeName, val code: Code)
+
+open class LibraryDefinition(
```
altavir, line 8: Why open class and not an interface with default methods? Also, this definition does not to create definitions directly from the code. We would like to create a Library definition as a class without and String code injections and use it directly. Maybe it is better to create a definition interface that uses functions instead of `Code`  and an abstract class that uses code blocks to implement those functions.

jupyter-lib/kotlin-jupyter-api/src/main/kotlin/org/jetbrains/kotlin/jupyter/api/results.kt
```diff
@@ -0,0 +1,55 @@
+package org.jetbrains.kotlin.jupyter.api
+
+import com.beust.klaxon.JsonObject
```
altavir, line 3: Why klaxon and not Serialization runtime? Serialization runtime is small and more widely used meaning that  you won't have to convert between different json representations.

altavir (commented):
Additionally to the comments above, I think that `use` should be a function of the notebook, so it could be called both by magics and by implementing and calling library descriptor manually.

jupyter-lib/kotlin-jupyter-api/src/main/kotlin/org/jetbrains/kotlin/jupyter/api/api.kt
ileasile: Added Javadoc, also added `addLibrary` method

jupyter-lib/kotlin-jupyter-api/src/main/kotlin/org/jetbrains/kotlin/jupyter/api/libraryConnection.kt
ileasile: Changed to something similar. Now definition should return the lists of `Execution` objects, they may be either simple functions or code snippets (instances of `CodeExecution` in the latter case)

jupyter-lib/kotlin-jupyter-api/src/main/kotlin/org/jetbrains/kotlin/jupyter/api/results.kt
ileasile: Migrated to kotlinx-serialization

PR #102, opened 2020-09-09
Allow REPL/Kernel to be embedded and share classes/variables with its caller

fmagin:
Currently the Kotlin kernel supports something similar to [IPython.start_kernel](https://ipython.readthedocs.io/en/stable/api/generated/IPython.html#IPython.start_kernel). This PR aims to support something that covers the use cases of [IPython.embed_kernel](https://ipython.readthedocs.io/en/stable/api/generated/IPython.html#IPython.embed_kernel).

#### What this PR provides

An application can now call `org.jetbrains.kotlin.jupyter.iKotlin.kernelServer` directly and pass a desired `KernelConfig` and the runtime properties where the `embedded` property is set to `true` and return the `KernelConfig` as a standard JSON to user in an application specific way.
The user can then run `jupyter console --existing=path/to/kernelconfig.json` to attach a Kotlin REPL to the kernel and run arbitrary Kotlin code inside the application context with the same state as the the application. 

#### What this PR does not allow

This does not allow attaching a Jupyter Notebook to the Kernel. This is a [limitation of Jupyter Notebooks](https://stackoverflow.com/questions/31382405/ipython-notebook-how-to-connect-to-existing-kernel/31385337#31385337) and not the Kotlin kernel.

### Status

This should work already, but I'm currently still writing the actual embedding code for the application I want to use it in, so until then I still consider this PR a draft, but appreciate an feedback on the general idea and current implementation.

EDIT:
The first proof of concept for this already works and I think anything else I wanted to achieve is an issue with the application embedding the kernel. This embedding could be simplified by allowing the embedding Java Application to acess `defaultRuntimeProperties`. I am also feeling unsure if the runtimeproperties is really the right place for the embedding option, and if `KernelArgs`/`KernelConfig` wouldn't be a better place.

### DIFF
--- a/src/main/kotlin/org/jetbrains/kotlin/jupyter/config.kt
+++ b/src/main/kotlin/org/jetbrains/kotlin/jupyter/config.kt
@@ -83,13 +83,13 @@ class RuntimeKernelProperties(val map: Map<String, String>): ReplRuntimeProperti
     }
 }
 
-data class KernelCfgFile(
+data class KernelJupyterParams(
         val sigScheme: String?,
         val key: String?,
         val ports: List<Int>,
         val transport: String?) {
     companion object {
-        fun fromFile(cfgFile: File): KernelCfgFile {
+        fun fromFile(cfgFile: File): KernelJupyterParams {
             val cfgJson = Parser.default().parse(cfgFile.canonicalPath) as JsonObject
             fun JsonObject.getInt(field: String): Int = int(field)
                     ?: throw RuntimeException("Cannot find $field in $cfgFile")
@@ -98,7 +98,7 @@ data class KernelCfgFile(
             val key = cfgJson.string("key")
             val ports = JupyterSockets.values().map { cfgJson.getInt("${it.name}_port") }
             val transport = cfgJson.string("transport") ?: "tcp"
-            return KernelCfgFile(sigScheme, key, ports, transport)
+            return KernelJupyterParams(sigScheme, key, ports, transport)
         }
     }
 }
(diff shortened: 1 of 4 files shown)


fmagin:
I have been testing and using this for the past week and it works quite well, and allows some really powerful extensions to at least one Java GUI tool that I am using. The approach should be general enough that this also works for every other JVM project that wants to simply embed a kernel in its context.

src/main/kotlin/org/jetbrains/kotlin/jupyter/config.kt
```diff
@@ -83,6 +83,26 @@ class RuntimeKernelProperties(val map: Map<String, String>): ReplRuntimeProperti
     }
 }
 
+data class KernelCfgFile(
```
ileasile, line 86: I think that name `KernelJupyterParams` fits better. Anyway, please rename this class to something without abbreviations

ileasile (commented):
I think that if it has real applications, this PR is OK, except one small thing. Please fix it and I'll merge it

src/main/kotlin/org/jetbrains/kotlin/jupyter/config.kt
fmagin: Sounds like a good name, changed it.

fmagin:
The testcases work for me locally, both `Test for Java 8` and `Test for Java 11 (set Gradle JVM to v.11)`. Not sure why the TeamCity build fails and I can't access the details.

PR #114, opened 2020-12-16
Add jDSP library to supported libraries.

biranyucel:


ileasile:
Thank you, @biranyucel!

PR #124, opened 2021-02-05
Support clients without stdin support (allow_stdin=false)

strangepleasures:


ileasile:
Thank you for your contribution!

Issue #128, opened 2021-02-23
Add Cell information to API render method


altavir:
I think the renderer should be available of its cell if possible, so it should be `inline fun <reified T : Any> render(noinline renderer: Cell.(T) -> Any)` instead of `inline fun <reified T : Any> render(noinline renderer: (T) -> Any)`.

ileasile:
You have an access to the `Notebook` instance, and you may simply refer this cell as `notebook?.currentCell`, why doesn't it suit you?

altavir:
Did not think about that. I think it would suffice. Still, from API wholeness point of view, it still makes sense. For example, it would be possible to create stand-alone renderers. Of course with `currentCell` it is just cosmetics.

PR #130, opened 2021-02-26
Move notebook from onLoaded to Builder argument

ileasile:


jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/libraries/JupyterIntegration.kt
```diff
@@ -69,9 +70,9 @@ abstract class JupyterIntegration : LibraryDefinitionProducer {
             fileAnnotations.add(handler)
         }
 
-        inline fun <reified T : Any> render(noinline renderer: (T) -> Any) {
+        inline fun <reified T : Any> render(noinline renderer: CodeCell.(T) -> Any) {
             val execution = ResultHandlerExecution { _, property ->
-                FieldValue(renderer(property.value as T), property.name)
+                FieldValue(renderer(notebook!!.currentCell!!, property.value as T), property.name)
```
altavir, line 75: Probably better to provide error message instead of generic NPE if notebook is not defined.

ileasile:
Fixes #128

ileasile:
This change is API-breaking, please update the API version in your lib

Issue #161, opened 2021-03-10
Kernel uses serialization runtime from itself, not the library


altavir:
```kotlin
@file:DependsOn("org.jetbrains.kotlinx:kotlinx-serialization-core-jvm:1.1.0")
import kotlinx.serialization.builtins.*
Boolean.serializer().descriptor.isInline
```

`isInline` property is available in serialization 1.1.0, but Jupyter can't find it. In the current version (`0.8.3.275`) serialization, 1.0.1 is used so it probably takes serialization from there.

altavir:
The latest build (`0.8.3.279`) Indeed fixes the problem, but it is only a workaround. Platform serialization should not leak into the notebook content.

ileasile:
Problem should be fixed in `0.8.3.280`
See commit adaadf2bbedf67a4560b14b59269703b96226497

PR #162, opened 2021-03-10
Update serialization version

altavir:
This is (probably) a temporary work-around for the problem described in #161. It forces the kernel to use the latest serialization.


ileasile:
Merged manually

PR #183, opened 2021-04-06
Allow value rendering using renderers processor from API

### LINKED ISSUE #182: Support rendering of an object dynamically inside CodeCell context

ileasile:
Fixes #182

Use it like this:
```kotlin
renderWithHost<C> { host, value: C -> notebook.renderersProcessor.renderValue(host, value_to_be_rendered) }
```


CommanderTvis (approved):

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/libraries/JupyterIntegration.kt
```diff
@@ -71,10 +71,14 @@ abstract class JupyterIntegration : LibraryDefinitionProducer {
         }
 
         inline fun <reified T : Any> render(noinline renderer: CodeCell.(T) -> Any) {
-            val execution = ResultHandlerExecution { _, property ->
+            return renderWithHost { _, value: T -> renderer(this, value) }
+        }
+
+        inline fun <reified T : Any> renderWithHost(noinline renderer: CodeCell.(ExecutionHost, T) -> Any) {
```
CommanderTvis, line 77: Just got that it is not what I need for #183. This function renders property; however, I need to render an arbitrary object.

CommanderTvis (changes requested):

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/TypeRenderersProcessor.kt
```diff
@@ -1,13 +1,12 @@
...
-import org.jetbrains.kotlinx.jupyter.api.Code
-import org.jetbrains.kotlinx.jupyter.api.FieldValue
-import org.jetbrains.kotlinx.jupyter.api.RendererTypeHandler
 import org.jetbrains.kotlinx.jupyter.api.libraries.ExecutionHost
 
 interface TypeRenderersProcessor {
 
     fun renderResult(host: ExecutionHost, field: FieldValue): Any?
 
+    fun renderValue(host: ExecutionHost, value: Any?): Any?
```
altavir, line 9: It is not clear what this function does. Documentation would be nice.

CommanderTvis (approved):

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/TypeRenderersProcessor.kt
altavir: I see a problem here. Suppose the value is a container (`List<Something>`). The type of actual element renderer could not be resolved in the runtime due to type erasure. Won't it require KType?

ileasile: Regarding documentation: I've added it.
Regarding renderers: yes, due to type erasure KClass is the most we can get in runtime. It is how all the rendering mechanics works, it only uses KClass for getting type information. We could add Ktype argument here, but it would require breaking the renderers API, and it will actually be useless because it cannot be used for cells results rendering.

PR #208, opened 2021-04-29
Update plotly descriptors

altavir:
* New package name
* version 0.4.0
* mavenCentral


ileasile:
Could you please make only one commit instead of a bunch and regenerate README (gradlew generateReadme)?

PR #215, opened 2021-04-30
Add support for variables declarations in library API

### LINKED ISSUE #206: Allow defining notebook values in the Notebook API
I would like to be able to define new global variables/values in the notebook both from library descriptors and the Notebook reflection API.

My current use case is the following: I want to be able to change the properties of the plotly server from inside the notebook. For example, I would like to start it, stop it, and change the port. Ideal way is to write something like this:
```kotlin
%use plotly-server

plotlyServer.stop()
plotlyServer.port  = 22
plotlyServer.start()
```
In order to do that In should be able to define new properties through the notebook API like `fun Builder.define(map: Map<String, Any>)` or `defineVariable(map: Map<String, Any>)`.

Current workaround suggested by @ileasile is to define empty variable via script execution:
```kotlin
val conf = execute("val libConf = LibConf(...); libConf").value as LibConf
```
and then substitute the value via 
```kotlin
onVariable<LibConf> { value, prop ->
    prop as KMutableProperty<*>
    // prop.set(..)
}
```

ileasile:
Fixes #206

There is a usage example in the test, @altavir, please review if it is OK

### DIFF
--- a/jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/KotlinKernelHost.kt
+++ b/jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/KotlinKernelHost.kt
@@ -34,4 +34,9 @@ interface KotlinKernelHost {
      * Adds a new library via its definition. Fully interchangeable with `%use` approach
      */
     fun addLibrary(library: LibraryDefinition)
+
+    /**
+     * Declares the given properties. They may be used in the notebook cells
+     */
+    fun declareProperties(properties: Iterable<PropertyDeclaration>)
 }
--- a/jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/fieldsHandling.kt
+++ b/jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/fieldsHandling.kt
@@ -3,6 +3,7 @@ package org.jetbrains.kotlinx.jupyter.api
 import kotlin.reflect.KClass
 import kotlin.reflect.KProperty
 import kotlin.reflect.KType
+import kotlin.reflect.full.createType
 import kotlin.reflect.full.isSubtypeOf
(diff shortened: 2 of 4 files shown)


jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/KotlinKernelHost.kt
```diff
@@ -34,4 +34,9 @@ interface KotlinKernelHost {
      * Adds a new library via its definition. Fully interchangeable with `%use` approach
      */
     fun addLibrary(library: LibraryDefinition)
+
+    /**
+     * Declares the given variables. They may be used in the notebook cells
+     */
+    fun declareVariables(variables: Iterable<VariableDeclaration>)
```
altavir, line 41: I think it is better to call them properties since they are not actually variable.

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/fieldsHandling.kt
```diff
@@ -36,3 +37,25 @@ class FieldHandlerByClass(
...
+        name,
+        value,
+        value?.let { it::class.starProjectedType } ?: Any::class.createType(nullable = true),
+        isMutable
+    )
+}
+
+fun KotlinKernelHost.declareVariable(variable: VariableDeclaration) = declareVariables(listOf(variable))
+fun KotlinKernelHost.declareVariable(name: VariableName, value: Any?) = declareVariable(VariableDeclaration(name, value))
+fun KotlinKernelHost.declareVariables(variables: Map<VariableName, Any?>) = declareVariables(variables.map { VariableDeclaration(it.key, it.value) })
```
altavir, line 61: Probably better to use `vararg pairs: Pair<VariableName, Any?>`. It would cover both this case and the line before.

altavir (commented):
Looks nice

PR #220, opened 2021-05-03
Add hierarchies visualization API to lib-ext

ileasile:
API for different hierarchies visualization.
Visualize class types hierarchy and classloaders hierarchy:
image

It is easy to define own wrappers for KMath expressions, KotlinDL network layers, Dataframe hierarchical keys, and so on.
Examples:
```kotlin
class HKClass(node: KClass<*>) : NodeWrapper<KClass<*>>(node) {
    override val ingoingEdges by lazy {
        node.superclasses.map { HKClass(it) }
    }
}

class HClassLoader(node: ClassLoader) : NodeWrapper<ClassLoader>(node) {
    override val ingoingEdges by lazy {
        node.parent?.let { listOf(HClassLoader(it)) } ?: emptyList()
    }
    override val label: String get() {
        return when (node) {
            is URLClassLoader -> node.urLs.joinToString("\\n", "URL ClassLoader:\\n") {
                it.toString()
            }
            else -> node.toString()
        }
    }
}
```

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -28,6 +28,6 @@ org.gradle.jvmargs=-XX:MaxMetaspaceSize=1024m -Xmx2048m
 # Turn off README check when running check task
 skipReadmeCheck=false
 
-jupyterApiVersion=0.9.1-42
+jupyterApiVersion=0.9.1-45
 kotlin.jupyter.add.api=false
 kotlin.jupyter.add.scanner=false
--- a/jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/graphs/GraphNode.kt
+++ b/jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/graphs/GraphNode.kt
@@ -1,0 +1,38 @@
+package org.jetbrains.kotlinx.jupyter.api.graphs
+
+/**
+ * Graph node which represents the object as a part of some hierarchy
+ *
+ * Classes implementing this interface should take care of [equals] and [hashCode]
+ * because they are used for testing the nodes for equality, and wrong implementation
+ * of these methods may lead to the wrong graph rendering, StackOverflow / OutOfMemory
+ * errors and so on. See example in [NodeWrapper]
+ *
+ * @param T Underlying object type
+ */
+interface GraphNode<T> {
+    /**
+     * Node label with all required information
+     */
+    val label: Label
+
+    /**
+     * Nodes which are connected with the ingoing edges to this one:
+     * {this} <- {inNode}
+     */
+    val inNodes: List<GraphNode<T>>
+
+    /**
+     * Nodes which are connected with the outgoing edges to this one:
+     * {this} -> {outNode}
+     */
+    val outNodes: List<GraphNode<T>>
+
+    /**
+     * Nodes which are connected with the undirected edges to this one:
+     * {this} -- {biNode}
+     */
+    val biNodes: List<GraphNode<T>>
+
+    companion object
+}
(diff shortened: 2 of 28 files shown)


jupyter-lib/lib-ext/src/main/kotlin/org/jetbrains/kotlinx/jupyter/ext/graph/wrappers/HClassLoader.kt
```diff
@@ -0,0 +1,18 @@
+package org.jetbrains.kotlinx.jupyter.ext.graph.wrappers
+
+import org.jetbrains.kotlinx.jupyter.ext.graph.NodeWrapper
+import java.net.URLClassLoader
+
+class HClassLoader(node: ClassLoader) : NodeWrapper<ClassLoader>(node) {
```
altavir, line 6: Not clear what H means. It is also better to create an extension function like `inline fun <reified T> GraphNode.Companion.fromClassLoader<T>(): GraphNode` for builders. It would greatly improve discoverability.

jupyter-lib/lib-ext/src/main/kotlin/org/jetbrains/kotlinx/jupyter/ext/graph/GraphNode.kt
```diff
@@ -0,0 +1,10 @@
+package org.jetbrains.kotlinx.jupyter.ext.graph
+
+interface GraphNode<T> {
+    val node: T
+    val label: String
+
+    val ingoingEdges: List<GraphNode<T>>
```
zaleslaw, line 7: Maybe incomingNodes? It will be consistent with the type

jupyter-lib/lib-ext/src/main/kotlin/org/jetbrains/kotlinx/jupyter/ext/graph/GraphNode.kt
```diff
@@ -0,0 +1,10 @@
...
+    val outgoingEdges: List<GraphNode<T>>
```
zaleslaw, line 8: Maybe outcomingNodes?

jupyter-lib/lib-ext/src/main/kotlin/org/jetbrains/kotlinx/jupyter/ext/graph/GraphNode.kt
```diff
@@ -0,0 +1,10 @@
+package org.jetbrains.kotlinx.jupyter.ext.graph
+
+interface GraphNode<T> {
+    val node: T
```
zaleslaw, line 4: Does graphNode contain the "node"? Do you mean additional information related to the Node? Unclear naming

jupyter-lib/lib-ext/src/main/kotlin/org/jetbrains/kotlinx/jupyter/ext/graph/GraphNode.kt
```diff
@@ -0,0 +1,10 @@
...
+    val label: String
```
zaleslaw, line 5: Maybe better to have generic here, for label type?

jupyter-lib/lib-ext/src/main/kotlin/org/jetbrains/kotlinx/jupyter/ext/graph/wrappers/HClassLoader.kt
ileasile: Good point, thanks!

jupyter-lib/lib-ext/src/main/kotlin/org/jetbrains/kotlinx/jupyter/ext/graph/GraphNode.kt
ileasile: WDYT about parentNodes/childNodes/siblingNodes? I think it's incorrect to say "incoming" about graph nodes, but I feel your concern is valid

PR #240, opened 2021-05-18
Add a possibility to define own implicit receivers

### LINKED ISSUE #239: Add notebook-wide implicit cell receivers
**The use case**
Consider the code here: https://github.com/AndreiKingsley/kmath-tensors-example-jupyters/blob/master/kmath_tensors_examples.ipynb. KMath extensively uses context-oriented programming so each section on the code should be encapsulated with the appropriate receiver. Usually, we just do something like `fun main() = DoubleTensorAlgebra{...}` to avoid additional clutter. In jupyter one have to wrap each cell in it which looks ugly.

Additional use cases:
* Implicit `CoroutineScope` for the notebook to allow convenient coroutines without explicit top-level scopes (and to avoid GlobalScope).
* Use aspects like logging (just add logger object as an implicit receiver).
* Prototype multiple receiver API without actually having this feature in the language.

**Proposed syntax**
```kotlin
/**
 * Add an implicit receiver to all subsequent cells. The `typeOf<T>` is used as a key for the stored receiver
 */
public inline fun <reified T: Any> Notebook.withReceiver(obj: T)

/**
 * Remove receiver with the appropriate key from subsequent cells
 */
public inline fun <reified T: Any> Notebook.removeReceiver()
```

Functions are used on `Notebook` to improve discoverability and avoid accidentally calling them.

**Risks**
Broad usage of implicit (including implicit receivers) could cause a lot of confusion in a large code, so this feature should not be used in large notebooks. The receiver should be declared in the notebook explicitly so it probably should not be used in plugins.

Adding multiple receivers which export the same functions and properties could produce unexpected behavior since the resolution could depend on the order. So there should be cell magic to list currently loaded receivers.

ileasile:
Fixes #239

### DIFF
--- a/src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/ReplWithReceiverTests.kt
+++ b/src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/ReplWithReceiverTests.kt
@@ -6,14 +6,41 @@ import org.jetbrains.kotlinx.jupyter.test.classpath
 import org.junit.jupiter.api.Test
 import java.io.File
 import kotlin.test.assertEquals
+import kotlin.test.assertFails
 
 class ReplWithReceiverTests : AbstractReplTest() {
     @Test
     fun testReplWithReceiver() {
         val value = 5
         val cp = classpath + File(ConstReceiver::class.java.protectionDomain.codeSource.location.toURI().path)
-        val repl = ReplForJupyterImpl(resolutionInfoProvider, cp, null, scriptReceivers = listOf(ConstReceiver(value)))
+        val repl = ReplForJupyterImpl(resolutionInfoProvider, cp, null, implicitReceivers = listOf(ConstReceiver(value)))
         val res = repl.eval("value")
         assertEquals(value, res.resultValue)
+    }
+
+    @Test
+    fun testReplWithAdHocReceivers() {
+        val repl = makeSimpleRepl()
+        repl.eval(
+            """
+            class A(val x: Int)
+            object B { val y = 42 }
+            """.trimIndent()
+        )
+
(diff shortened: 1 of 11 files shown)


altavir (approved):

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/ReplWithReceiverTests.kt
```diff
@@ -6,14 +6,41 @@ import org.jetbrains.kotlinx.jupyter.test.classpath
...
+        repl.eval(
+            """
+            class A(val x: Int)
+            object B { val y = 42 }
+            """.trimIndent()
+        )
+
+        repl.eval(
+            """
+            HOST.withReceiver(A(50))
```
altavir, line 33: Still look somewhat strange. Not everybody understands what `host` is.

ileasile: Maybe `executor`?

altavir: From the logical point of view, the receiver is applied to cells. So the code in cells is wrapped in the receiver. This is why I initially proposed to attach it to the notebook. It seems like a non-important thing, but discoverability is important.

ileasile:
Postponed till the resolution of this script codegen bug: https://youtrack.jetbrains.com/issue/KT-47755

ileasile:
It turned out that removing of implicit receivers is beyond the functional abilities of the current scripting implementation. So, this PR may be merged only with the adding of the receivers

PR #287, opened 2021-06-28
Create specialized kernels w/ JDK, JVM args, and environment variables

### LINKED ISSUE #72: Pass args on JVM startup
Hi ! 

I'm hitting the 8G memory limit and wanted to pass the `-Xmx` startup flag but couldn't find how to it. I believe it's useful to be able to customize how the JVM starts up.
I'm not familiar with Jupyter's internals, but is this where the args should be passed ? 
https://github.com/Kotlin/kotlin-jupyter/blob/3e3ebaa0e47ac885f41552583b12b9110bd40740/distrib/run_kotlin_kernel/run_kernel.py#L42

Or should we pass them somewhere else ?

Cheers ! 

PS : If you have any tips on how to change the memory limit, I'm still interested :)

### LINKED ISSUE #284: Allow setting JDK to use
I have multiple JDKs installed, and I would like to be able to control which one my notebook runs with.  Ideally this would be a per-notebook setting, but a per-jupyter instance setting would also work.  `JAVA_HOME` does not seem to be used, so even just using that would be a significant upgrade.

If it's as simple as it looks (using JAVA_HOME [here](https://github.com/Kotlin/kotlin-jupyter/blob/3e3ebaa0e47ac885f41552583b12b9110bd40740/distrib/run_kotlin_kernel/run_kernel.py#L42)) I can make a PR.

rnett:
Fixes #284.
Fixes #72.

I'd like to note this somewhere in the readme (and that it can be set per-kernel with `env` in the kernelspec), any preference on where?

I also thought about a few possible further features:
1. Creating kernels for each detected JDK version on install (presumably in `install_user`, although I can't find where it gets called on initial install).
2. Adding a `fix-kernelspec-location`-like task to create a kernel for a specific JDK.
3. Adding a `kotlin-jupyter` specific environment variable that overrides `JAVA_HOME`, i.e. `KOTLIN_JUPYTER_JAVA`, incase you want jupyter on a different JDK than say gradle (although that's doable via the kernelspec).

Thoughts on any of them?


ileasile:
Hi!
To satisfy CI you may rebase on the latest master
You may mention this feature in `Usage` section in README

Regarding your questions:
1. Good idea, it may be done in setup script, I suppose
2. It's also worth adding
3. Agree, and it may be even done in this PR.

rnett:
Autodetection is on hold for now since it's non-trivial, I'd probably do it in another PR.

The rest is done, although I did use `subprocess.check_output` which is from Python 3.1.  As far as I can tell you're targeting 3, should I use `subprocess.PIPE` or is it a non-issue?

rnett:
Also, is there a reason `install_user` isn't called from the setup script?

rnett:
Making this a draft for now, since I was working on #72 and think it's better done as a `create_kernel` command that can set any of the JDK, JVM args, or environment variables.

rnett:
Went faster than I expected, this fixes #72 too now.

docs/README.md
```diff
@@ -8,7 +8,7 @@
 
 # Kotlin kernel for IPython/Jupyter
 
-[Kotlin](https://kotlinlang.org/) (1.5.30-dev-2630) kernel for [Jupyter](https://jupyter.org).
+[Kotlin](https://kotlinlang.org/) (1.5.20) kernel for [Jupyter](https://jupyter.org).
```
ileasile, line 11: Shouldn't be changed by regeneration, I suppose. Please, rebase on master and use default Kotlin version, dependencies problems should be fixed now

distrib/kotlin_kernel/constants.py
```diff
@@ -0,0 +1,17 @@
+# standard JVM options environment variable
```
ileasile, line 1: `constants` is too generic name. Maybe `env_names` or `envs`, what do you think?

ileasile (commented):
Great work, @rnett, thank you A LOT!
I have some minor comments. After fixing I'll publish dev packages from your branch to conda and pip so we could test them.

ileasile:
>The rest is done, although I did use subprocess.check_output which is from Python 3.1. As far as I can tell you're targeting 3, should I use subprocess.PIPE or is it a non-issue?
It's ok. I think, 3.4 is the minimal version we should support

distrib/kotlin_kernel/constants.py
rnett: Build works fine for me now, and I like `env_names`.

ileasile:
@rnett Could you please install and try the kernel built from this PR? If everything is fine, I'll merge it
```pip install -i https://test.pypi.org/simple/ kotlin-jupyter-kernel==0.10.0.83.dev1```
or
```conda install -c jetbrains-dev kotlin-jupyter-kernel==0.10.0.83.dev1```

rnett:
These was a typo in the readme, but other than that everything worked.  Tested environment variables and kernel creation on Windows and WSL.

ileasile:
Thank you, Ryan!

PR #297, opened 2021-07-06
Variables view feature

nikolay-egorov:


src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
```diff
@@ -124,4 +158,73 @@ internal class InternalEvaluatorImpl(
...
+
+            if (state.stringValue != newValue.toString()) {
+                usageMap[cellId]?.add(it.key)
+                state.stringValue = newValue.toString()
+            }
+        }
+    }
+
+    private fun getVisibleVariables(target: ResultValue, cellId: Int): Map<String, VariableState> {
+        val klass = target.scriptClass ?: return emptyMap()
```
ileasile, line 181: Please use `kClass` or `class` word in naming

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
```diff
@@ -124,4 +158,73 @@ internal class InternalEvaluatorImpl(
...
+        val cellKlassInstance = target.scriptInstance!!
+
+        val fields = klass.declaredMemberProperties
+        val ans = mutableMapOf<String, VariableState>()
+        fields.forEach {
+            val casted = it as KProperty1<Any, *>
```
ileasile, line 187: There is no word `casted`. Moreover, if you simply write `it as T`, `it` will have type `T` in the following lines due to smartcasting

ileasile: But I suggest to use named lambda parameter here

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
```diff
@@ -124,4 +158,73 @@ internal class InternalEvaluatorImpl(
...
+            ans[casted.name] = if (casted.get(cellKlassInstance) == null) {
+                return@forEach
```
ileasile, line 189: What about changing `.forEach`-`return@forEach` to simple `for`-`continue`?

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
```diff
@@ -124,4 +158,73 @@ internal class InternalEvaluatorImpl(
...
+
+    private fun getVisibleVariables(target: ResultValue, cellId: Int): Map<String, VariableState> {
+        val klass = target.scriptClass ?: return emptyMap()
+        val cellKlassInstance = target.scriptInstance!!
+
+        val fields = klass.declaredMemberProperties
+        val ans = mutableMapOf<String, VariableState>()
+        fields.forEach {
+            val casted = it as KProperty1<Any, *>
+            ans[casted.name] = if (casted.get(cellKlassInstance) == null) {
```
ileasile, line 188: Why variable value can't be null?

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
```diff
@@ -124,4 +158,73 @@ internal class InternalEvaluatorImpl(
...
+                return@forEach
+            } else {
+                val stringVal = casted.get(cellKlassInstance).toString()
```
ileasile, line 191: Calling `.get()` may be expensive, it's better to avoid making it twice

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
```diff
@@ -124,4 +158,73 @@ internal class InternalEvaluatorImpl(
...
+                // redeclaration of any type
+                if (containingMap.containsKey(casted.name)) {
+                    val oldCellId = containingMap[casted.name]
+                    if (oldCellId != cellId) {
+                        usageMap[oldCellId]?.remove(casted.name)
+                    }
+                }
+                // it was val, now it's var
+                if (it is KMutableProperty1 && variablesCache.containsKey(casted.name)) {
+                    variablesCache.remove(casted.name)
```
ileasile, line 203: What if we call `remove` on absent key? I think, this condition may be simplified

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
```diff
@@ -57,6 +66,17 @@ internal class InternalEvaluatorImpl(
 
     override val lastClassLoader get() = compiler.lastClassLoader
 
+    override val variablesMap = mutableMapOf<String, VariableState>()
```
ileasile, line 69: Difference between variablesCache and variablesState is not clear

src/main/kotlin/org/jetbrains/kotlinx/jupyter/util.kt
```diff
@@ -10,6 +11,73 @@ import kotlin.script.experimental.jvm.util.toSourceCodePosition
 
 fun List<String>.joinToLines() = joinToString("\n")
 
+// todo : perhaps, create factory class
```
ileasile, line 14: Also, if you add TODOs, follow repository standard:
```
// TODO: <>
```

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/ReplTests.kt
```diff
@@ -440,4 +441,214 @@ class ReplTests : AbstractSingleReplTest() {
 
         assertEquals("org.RDKit.RWMol", res!!::class.qualifiedName)
     }
+
+    @Test
```
ileasile, line 445: If you add so much tests, you may even move them to a separate class/file

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/testUtil.kt
```diff
@@ -31,6 +32,25 @@ import org.jetbrains.kotlinx.jupyter.libraries.parseLibraryDescriptors
 import org.jetbrains.kotlinx.jupyter.log
 import org.jetbrains.kotlinx.jupyter.repl.CompletionResult
 import java.io.File
+import kotlin.collections.Collection
```
ileasile, line 35: `kotlin.collections` is a default import and is not needed here, it's interesting how it was added (maybe it's IDEA or Kotlin plugin bug, then it may be reported)

ileasile (changes requested):

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/Notebook.kt
```diff
@@ -11,6 +9,17 @@ interface Notebook {
...
+    /**
+     * Current state of visible variables
+     */
+    val variablesMap: MutableMap<String, VariableState>
+
+    /**
+     * Stores useful variables in a cell.
+     * Useful <==> declarations + modifying references
+     */
+    var usageMap: Map<Int, Set<String>>
```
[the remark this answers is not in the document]
nikolay-egorov: Yeah, would add more info

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/VariableState.kt
```diff
@@ -0,0 +1,9 @@
+package org.jetbrains.kotlinx.jupyter.api
+
+import kotlin.reflect.KProperty1
+
+data class VariableState(
+    val propertyKlass: KProperty1<Any, *>,
+    val scriptInstance: Any,
+    var stringValue: String
```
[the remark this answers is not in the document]
nikolay-egorov: So we basically would like to consider that VariableState is an immutable thing, hence, from the semantic perspective, changing stringValue gives us a new VariableState? Sure thing, I can have another var field in a implementation

src/main/kotlin/org/jetbrains/kotlinx/jupyter/apiImpl.kt
```diff
@@ -132,6 +136,23 @@ class NotebookImpl(
     override val jreInfo: JREInfoProvider
         get() = JavaRuntime
 
+    fun updateVarsState(varsMap: Map<String, VariableState>) {
+        variablesMap += varsMap
```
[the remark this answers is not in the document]
nikolay-egorov: Yup, very convenient feature!

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
```diff
@@ -1,24 +1,10 @@
...
-import org.jetbrains.kotlinx.jupyter.api.Renderable
-import org.jetbrains.kotlinx.jupyter.codegen.ClassAnnotationsProcessor
-import org.jetbrains.kotlinx.jupyter.codegen.ClassAnnotationsProcessorImpl
-import org.jetbrains.kotlinx.jupyter.codegen.FieldsProcessor
-import org.jetbrains.kotlinx.jupyter.codegen.FieldsProcessorImpl
-import org.jetbrains.kotlinx.jupyter.codegen.FileAnnotationsProcessor
-import org.jetbrains.kotlinx.jupyter.codegen.FileAnnotationsProcessorImpl
-import org.jetbrains.kotlinx.jupyter.codegen.ResultsRenderersProcessor
-import org.jetbrains.kotlinx.jupyter.codegen.RenderersProcessorImpl
+import org.jetbrains.kotlinx.jupyter.api.*
```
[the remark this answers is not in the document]
nikolay-egorov: Hmm, strange thing ktlFormat passed

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
```diff
@@ -395,7 +380,20 @@ class ReplForJupyterImpl(
...
+            printVars()
+//            printUsage(jupyterId, usageMap[jupyterId - 1]!!)
+
+            //todo: perhaps redundant
+            // would send serializable version
+            val mapToSend = mutableMapOf<String, String>()
+            varsMap.forEach {
+                val value = it.value.stringValue
+                mapToSend[it.key] = value
+            }
```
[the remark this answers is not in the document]
nikolay-egorov: Wow, useful map method

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
```diff
@@ -370,13 +349,19 @@ class ReplForJupyterImpl(
 
             val compiledData: SerializedCompiledScriptsData
             val newImports: List<String>
+            val varsMap : Map<String, VariableState>
+            val usageMap: Map<Int, Set<String>>
```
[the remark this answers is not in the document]
nikolay-egorov: We could. I was just following the intention of declaring them before try, just like with compiledData and newImports

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/Notebook.kt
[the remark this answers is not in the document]
nikolay-egorov: Do you agree to let it be var?

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
```diff
@@ -395,7 +380,20 @@ class ReplForJupyterImpl(
                 updateClasspath()
             } ?: emptyList()
 
-            EvalResult(rendered, EvaluatedSnippetMetadata(newClasspath, compiledData, newImports))
+            notebook.updateVarsState(varsMap)
+            notebook.usageMap = usageMap
+            printVars()
+//            printUsage(jupyterId, usageMap[jupyterId - 1]!!)
```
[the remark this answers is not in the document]
nikolay-egorov: What's it? You mean [this](https://kotlinlang.org/docs/coding-conventions.html#indentation)?

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
```diff
@@ -370,13 +349,19 @@ class ReplForJupyterImpl(
 
             val compiledData: SerializedCompiledScriptsData
             val newImports: List<String>
+            val varsMap : Map<String, VariableState>
+            val usageMap: Map<Int, Set<String>>
+            internalEvaluator.lastExecutedCellId = jupyterId - 1
```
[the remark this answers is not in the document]
nikolay-egorov: Yeah, we can send it in the call back, like this: 
`                executor.execute(code, displayHandler) { jupyterId, codeToExecute ->`

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
```diff
@@ -57,6 +66,17 @@ internal class InternalEvaluatorImpl(
...
+
+    override val usageMap = mutableMapOf<Int, MutableSet<String>>()
+
+    override var lastExecutedCellId: Int = 0
+
+    /**
+     * Tells in which cell a variable was declared
+     */
+    private val containingMap: MutableMap<String, Int> = mutableMapOf()
```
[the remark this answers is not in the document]
nikolay-egorov: Alright, it might be a kotlin implicit types thing in usages, that's why I kept carrying on

src/main/kotlin/org/jetbrains/kotlinx/jupyter/util.kt
```diff
@@ -10,6 +11,73 @@ import kotlin.script.experimental.jvm.util.toSourceCodePosition
...
+fun generateHTMLVarsReport(variablesMap: Map<String, VariableState>): String {
+    val stringBuilder = StringBuilder(
```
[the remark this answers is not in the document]
nikolay-egorov: Interesting, would do

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
```diff
@@ -124,4 +158,73 @@ internal class InternalEvaluatorImpl(
...
+                    return@forEach
+                }
+                state
+            }
+        }
+        return ans
+    }
+
+    // this is the O(n) implementation
+    @Deprecated("Back-propagation def search", ReplaceWith("updateVariablesState"))
```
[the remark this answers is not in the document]
nikolay-egorov: Hmmm)

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
nikolay-egorov: Seems like a naming convention, doesn't it?

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
nikolay-egorov: Oh, yeah, forgot about that casting context changer. And named parameter is also cool

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
nikolay-egorov: Let's discuss! From what I know about Java, all these lambdas are stored in an inner class inside a calling class. Yes, it may sound like a solution with overhead to resolve needed one and call a method, but I rely on the fact we do it repeatedly. 
So, as it seems to me, both these solutions are equal, being the only difference that lambda with that `return-to-label` being a bit fancier

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
nikolay-egorov: Fair enough)

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
nikolay-egorov: Perhaps, I should add more comments? The idea was the following: 
-variablesCache stores info about vals 
-variablesMap about everything 
so first is a part of the second, and there is an invariant for them

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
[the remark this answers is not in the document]
nikolay-egorov: Now it's official - ktlintFormat is satisfied with this file

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
nikolay-egorov: I'm thinking about this invariant: all process is here, so we can literally send "null" as a string value. Would it be an overhead? Instead, we could send just String -> String?, and not overload the channel with needless bytes

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/InternalEvaluator.kt
```diff
@@ -12,6 +13,13 @@ interface InternalEvaluator {
     val lastKClass: KClass<*>
     val lastClassLoader: ClassLoader
 
+    val variablesMap: MutableMap<String, VariableState>
+
+    val usageMap: MutableMap<Int, MutableSet<String>>
+
+    // todo: perhaps, better to be stateless
+    var lastExecutedCellId: Int
```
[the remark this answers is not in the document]
nikolay-egorov: Or maybe we get rid of this if we send it as a parameter in a executionCallBackl

src/main/kotlin/org/jetbrains/kotlinx/jupyter/util.kt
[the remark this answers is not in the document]
nikolay-egorov: BTW, I did not get the fun of `buildString` func. It seems just like calling SB methods from lambda and then returning `toString()`
And I'm also curious about compiler optimisations of these types of injections if we call generation multiple times

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/VariableState.kt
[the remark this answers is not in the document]
ileasile: Basically, property class and script instance hold the exhaustive information about variable state. Other things may be obtained from these two. So I suggest something like this:
```kotlin
data class VariableState(
    private val propertyClass: KProperty1<Any, *>,
    private val scriptInstance: Any,
) {
    // caching, methods to get pure and stringified values go here
}
```
What do you think?

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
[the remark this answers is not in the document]
ileasile: I mean that commented line should have the same indent as the line above)
```
    printVars()
    // printUsage(jupyterId, usageMap[jupyterId - 1]!!)
```

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/VariableState.kt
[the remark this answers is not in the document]
nikolay-egorov: Cool, I did this way : 
```
interface VarState {
    val propertyKlass: KProperty<*>
    val scriptInstance: Any?
    val stringValue: String?
}

data class VariableState(
    override val propertyKlass: KProperty1<Any, *>,
    override val scriptInstance: Any,
    var mutableStringValue: String?
) : VarState {
    override val stringValue: String?
        get() = mutableStringValue
}
```
But I like your aproach more

src/main/kotlin/org/jetbrains/kotlinx/jupyter/util.kt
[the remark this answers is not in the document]
ileasile: It should be optimized, yes

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
ileasile: If you have any references, please share. Common replacement of `class` is `clazz`, but if it's a part of the identifier, you don't have to replace it.

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
ileasile: I'm not talking about performance) If you really love this `return@forEach` construction, leave it as is, but I'd use for-continue here

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
ileasile: I think that "null" is a legit value for user to see. I don't think that we should optimize something here

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/testUtil.kt
nikolay-egorov: Yes, it's a strange thing - I removed it and neither Idea's imports optimisation nor ktlint gave it back

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
nikolay-egorov: No, I do not have references, just tried to name class) But stick with a kClass

src/main/kotlin/org/jetbrains/kotlinx/jupyter/apiImpl.kt
```diff
@@ -132,6 +136,23 @@ class NotebookImpl(
     override val jreInfo: JREInfoProvider
         get() = JavaRuntime
 
+    fun updateVarsState(varsMap: Map<String, VariableState>) {
```
ileasile, line 139: `varsMap` and `variablesMap` are semantically same things, but in this context they mean different entities. `varsMap` is a "state update", but `variablesMap` is a "whole state". Please rethink the naming you're using. Naming should be a reflection of variable meaning and of the reason it was created, not just a way of referencing

src/main/kotlin/org/jetbrains/kotlinx/jupyter/apiImpl.kt
```diff
@@ -132,6 +136,23 @@ class NotebookImpl(
...
+        variablesMap += varsMap
+    }
+
+    fun varsAsHtmlTable(): String {
+        return generateHTMLVarsReport(variablesMap)
+    }
+
+    fun varsAsString(): String {
```
ileasile, line 147: Maybe `variablesReport`? It's not just stringified variable values, it's a report, with a header and variable names

src/main/kotlin/org/jetbrains/kotlinx/jupyter/apiImpl.kt
```diff
@@ -132,6 +136,23 @@ class NotebookImpl(
...
+        if (variablesMap.isEmpty()) return ""
+        val stringBuilder = StringBuilder("Visible vars: \n")
+        variablesMap.forEach { (t, u) ->
+            stringBuilder.append("\t$t : ${u.stringValue}\n")
+        }
+        return stringBuilder.append('\n').toString()
```
ileasile, line 153: Use `buildString` to save a couple of characters. You may use saved characters for specifying understandable names for `t` and in `u` in `forEach`

src/main/kotlin/org/jetbrains/kotlinx/jupyter/htmlUtil.kt
```diff
@@ -0,0 +1,76 @@
+package org.jetbrains.kotlinx.jupyter
+
+import org.jetbrains.kotlinx.jupyter.api.VariableState
+
+// TODO : perhaps, create factory class
+fun generateHTMLVarsReport(variablesMap: Map<String, VariableState>): String {
```
ileasile, line 6: Same thing about naming. Please find all the usages of `variablesMap` and `varsMap` and think about better names

src/main/kotlin/org/jetbrains/kotlinx/jupyter/htmlUtil.kt
```diff
@@ -0,0 +1,76 @@
...
+    return buildString {
+        this.append(
+            """
+            <!DOCTYPE html>
+            <html>
```
ileasile, line 11: Do we need a full HTML document here, isn't simple `div` enough? Just wondering

src/main/kotlin/org/jetbrains/kotlinx/jupyter/htmlUtil.kt
```diff
@@ -0,0 +1,76 @@
+package org.jetbrains.kotlinx.jupyter
+
+import org.jetbrains.kotlinx.jupyter.api.VariableState
+
+// TODO : perhaps, create factory class
+fun generateHTMLVarsReport(variablesMap: Map<String, VariableState>): String {
+    return buildString {
+        this.append(
```
ileasile, line 8: You don't have to use `this` here

src/main/kotlin/org/jetbrains/kotlinx/jupyter/htmlUtil.kt
```diff
@@ -0,0 +1,76 @@
...
+            return this.toString()
+        }
+
+        this.append(generateVarsTable(variablesMap))
+
+        this.append("</body>\n</html>")
+    }
+}
+
+// TODO: text is not aligning in a center
```
ileasile, line 32:
```suggestion
// TODO: text is not aligned in the center
```

ileasile: By the way, you may spend a bit of time to get rid of this todo, or we may look together into it

src/main/kotlin/org/jetbrains/kotlinx/jupyter/htmlUtil.kt
```diff
@@ -0,0 +1,76 @@
...
+      padding: ${paddingPx}px;
+    }
+    </style>
+    """.trimIndent()
+    return styleSection
+}
+
+fun generateVarsTable(variablesMap: Map<String, VariableState>): String {
+    return buildString {
+        this.append(
```
ileasile, line 52: Same thing about `this`

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
```diff
@@ -360,6 +362,22 @@ class ReplForJupyterImpl(
         else context.compilationConfiguration.asSuccess()
     }
 
+    /**
+     * Used for debug purposes.
+     * @see ReplCommand
+     */
+    private fun printVars(isHtmlFormat: Boolean = false) = log.debug(
```
ileasile, line 369: Let's don't use `vars` word, only `variables`?

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
```diff
@@ -371,13 +389,17 @@ class ReplForJupyterImpl(
...
             val result = try {
-                executor.execute(code, displayHandler) { internalId, codeToExecute ->
+                log.debug("Current cell id: $jupyterId")
+                executor.execute(code, displayHandler, currentCellId = jupyterId - 1) { internalId, codeToExecute ->
                     cell = notebook.addCell(internalId, codeToExecute, EvalData(jupyterId, code))
                 }
             } finally {
                 compiledData = internalEvaluator.popAddedCompiledScripts()
                 newImports = importsCollector.popAddedImports()
+
```
ileasile, line 399: Remove this empty line

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
```diff
@@ -371,13 +389,17 @@ class ReplForJupyterImpl(
...
             }
+            val varsMap = internalEvaluator.variablesHolder
+            val usageMap = internalEvaluator.varsUsagePerCell
```
ileasile, line 402: Naming

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
```diff
@@ -395,7 +417,15 @@ class ReplForJupyterImpl(
...
             } ?: emptyList()
 
-            EvalResult(rendered, EvaluatedSnippetMetadata(newClasspath, compiledData, newImports))
+            notebook.updateVarsState(varsMap)
+            notebook.usageMap = usageMap
+            // printVars()
+            // printUsage(jupyterId, usageMap[jupyterId - 1]!!)
+
+            // TODO: perhaps redundant
+            // send serializable version
```
ileasile, line 426: Let's remove these comments

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/CellExecutor.kt
```diff
@@ -17,6 +17,7 @@ interface CellExecutor : ExecutionHost {
         processAnnotations: Boolean = true,
         processMagics: Boolean = true,
         invokeAfterCallbacks: Boolean = true,
+        currentCellId: Int = 1,
```
ileasile, line 20: First cell has `ID = 0`, isn't it? If so, `0` is a better default value. Or even `-1` to better distinguish cases of set and not set value

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/InternalEvaluator.kt
```diff
@@ -12,11 +13,15 @@ interface InternalEvaluator {
...
+    val variablesHolder: MutableMap<String, VariableState>
+
+    val varsUsagePerCell: MutableMap<Int, MutableSet<String>>
+
     /**
      * Executes code snippet
      * @throws IllegalStateException if this method was invoked recursively
      */
-    fun eval(code: Code, onInternalIdGenerated: ((Int) -> Unit)? = null): InternalEvalResult
+    fun eval(code: Code, cellId: Int = 1, onInternalIdGenerated: ((Int) -> Unit)? = null): InternalEvalResult
```
ileasile, line 24: Same thing about default value

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/InternalEvaluator.kt
```diff
@@ -12,11 +13,15 @@ interface InternalEvaluator {
     val lastKClass: KClass<*>
     val lastClassLoader: ClassLoader
 
+    val variablesHolder: MutableMap<String, VariableState>
+
+    val varsUsagePerCell: MutableMap<Int, MutableSet<String>>
```
ileasile, line 18: `vars` -> `variables`

ileasile, line 18: I don't think we need mutability here, in interface

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
```diff
@@ -31,6 +35,12 @@ internal class InternalEvaluatorImpl(
 
     private val registeredCompiledScripts = arrayListOf<SerializedCompiledScript>()
 
+    /**
+     * Stores cache for val values.
+     * Its contents is a part of variablesHolder.
+     */
+    private val variablesCache = mutableMapOf<String, VariableState>()
```
ileasile, line 42: Comment made the things clearer. So, it may be called `valVariablesCache`. Or it also may be a set of names, right? I don't see the reason to store same references in several maps

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
```diff
@@ -57,9 +67,18 @@ internal class InternalEvaluatorImpl(
 
     override val lastClassLoader get() = compiler.lastClassLoader
 
+    override val variablesHolder = mutableMapOf<String, VariableState>()
+
+    override val varsUsagePerCell = mutableMapOf<Int, MutableSet<String>>()
```
ileasile, line 72: -> variables

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
```diff
@@ -57,9 +67,18 @@ internal class InternalEvaluatorImpl(
...
+
+    /**
+     * Tells in which cell a variable was declared
+     */
+    private val varsDeclarationInfo: MutableMap<String, Int> = mutableMapOf()
```
ileasile, line 77: `varsUsagePerCell` and `varsDeclarationInfo` is actually a bidirectional map, right? You may write a generic utility class that incapsulates the logic of working with them.
By the way, you've added 4 properties here for a single feature. It would be great if their number might be reduced (i.e. using incapsualtion to another class(es))

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
```diff
@@ -124,4 +146,69 @@ internal class InternalEvaluatorImpl(
...
+            varsDeclarationInfo[it] != cellId
+        }
+
+        variablesHolder.forEach {
+            val state = it.value
+            val oldValue = state.stringValue
+            state.update()
+
+            if (state.stringValue != oldValue) {
+                varsUsagePerCell[cellId]?.add(it.key)
```
ileasile, line 162: So, we store modifications here.. Why is it "usage" map then? And how do we use this map, why is it needed for?

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
```diff
@@ -124,4 +146,69 @@ internal class InternalEvaluatorImpl(
...
+            }
+        }
+    }
+
+    private fun getVisibleVariables(target: ResultValue, cellId: Int): Map<String, VariableState> {
+        val kClass = target.scriptClass ?: return emptyMap()
+        val cellKlassInstance = target.scriptInstance!!
```
ileasile, line 169:
```suggestion
        val cellClassInstance = target.scriptInstance!!
```

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/ReplTests.kt
```diff
@@ -441,3 +442,217 @@ class ReplTests : AbstractSingleReplTest() {
...
+
+        val setOfCell = setOf("z", "f", "x")
+        assertTrue(state.containsValue(setOfCell))
+    }
+
+    @Test
+    fun testSeparateDefsUsage() {
+        eval(
+            """
+            val x = "abcd"
```
ileasile, line 604: Please make the tests for private/protected/internal variables:
```
private val x = 5
```
We should be sure that such declarations do not crash the notebook

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/CustomLibraryResolverTests.kt
```diff
@@ -244,7 +244,7 @@ class CustomLibraryResolverTests : AbstractReplTest() {
             %use lib1
             5
         """.trimIndent()
-        repl.execute(code)
+        repl.execute(code = code)
```
ileasile, line 247: Is it required change?

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/IntegrationApiTests.kt
```diff
@@ -104,7 +104,7 @@ class IntegrationApiTests {
         }
         val repl = makeRepl(lib).trackExecution()
 
-        repl.execute("%use mylib\n1")
+        repl.execute(code = "%use mylib\n1")
```
ileasile, line 107: And this

ileasile (changes requested):

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/VariableState.kt
```diff
@@ -0,0 +1,24 @@
...
+    val propertyKlass: KProperty<*>
+    val scriptInstance: Any?
+    val stringValue: String?
+}
+
+data class VariableState(
+    override val propertyKlass: KProperty1<Any, *>,
+    override val scriptInstance: Any,
+) : VarState {
+    private var _cachedValue: String? = null
```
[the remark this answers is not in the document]
nikolay-egorov: Alright, saw this approach here in C++ 
https://kotlinlang.org/docs/coding-conventions.html#names-for-backing-properties

jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/compiler/util/serializedCompiledScript.kt
```diff
@@ -25,6 +25,7 @@ class EvaluatedSnippetMetadata(
     val newClasspath: Classpath = emptyList(),
     val compiledData: SerializedCompiledScriptsData = SerializedCompiledScriptsData.EMPTY,
     val newImports: List<String> = emptyList(),
+    val variablesMap: Map<String, String?> = mutableMapOf()
```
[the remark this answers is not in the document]
nikolay-egorov: I like `recievedVariabesState`, what'd you say?

src/main/kotlin/org/jetbrains/kotlinx/jupyter/apiImpl.kt
nikolay-egorov: Fair's fair

src/main/kotlin/org/jetbrains/kotlinx/jupyter/htmlUtil.kt
nikolay-egorov: Oh...

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
nikolay-egorov: Vars part of me hardly accepted the proposition, but accepted nevertheless

jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/compiler/util/serializedCompiledScript.kt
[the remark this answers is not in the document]
nikolay-egorov: Or evaluatedVariablesState

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
nikolay-egorov: "Usage" was motivated by the means that we stored vars that were used in modifying behaviour. We might use it to output only "interesting" vars per cell to the user

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
nikolay-egorov: Yes, seems like we might go for reduction here

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/CustomLibraryResolverTests.kt
nikolay-egorov: I was facing a strange bug with "method not found", but now it's gone, so no, it's not required

src/main/kotlin/org/jetbrains/kotlinx/jupyter/apiImpl.kt
nikolay-egorov: Yeah, seems like I have a struggle not to think of better naming once a task is done. Algorithms contents flashbacks be like

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/CellExecutor.kt
nikolay-egorov: -1 seems potentially more useful, yeah

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
nikolay-egorov: It is indeed. Sounds like an interesting thing, would try

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/ReplTests.kt
nikolay-egorov: Improved realisation. All working correctly, but we have a limitation such that it's not possible to access any fields with non-default modificators in separate cells to modify. 
For instance, we have 2 cells. 
One: 
`private val x = 5 `
Another:
```
private var z = 47
z += x
```
Compile-time error, I believe it's expected

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
nikolay-egorov: Although it's not a bidirectional map in its original sense. We do want to have a supplementary map of declarations, so it's more like String -> Int with Int ->Set<String>, hence, types are not symmetric. It took me a while but in a meanwhile I've got the bidirectional generic map))

src/main/kotlin/org/jetbrains/kotlinx/jupyter/htmlUtil.kt
nikolay-egorov: Moreover, we do not need div. It's generated by the default for the output HTML and builts table in

src/main/kotlin/org/jetbrains/kotlinx/jupyter/util.kt
```diff
@@ -68,3 +68,43 @@ fun Int.toSourceCodePositionWithNewAbsolute(code: SourceCode, newCode: SourceCod
 fun ResultsRenderersProcessor.registerDefaultRenderers() {
     register(bufferedImageRenderer)
 }
+
+/**
+ * Stores info about where a variable Y was declared and info about what are they at the address X.
+ * T: key, stands for a way of addressing variables, e.g. address.
+ * V: value, from Variable, choose any suitable type for your variable reference.
+ * Default: Int, String
```
ileasile, line 76: T = Int, V = String

ileasile: BTW, key type is usually named `K`, not `T`

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/ReplTests.kt
```diff
@@ -439,3 +441,335 @@ class ReplTests : AbstractSingleReplTest() {
...
+
+        val varsUpdate = mutableMapOf<String, String>(
+            "x" to "1",
+            "y" to "0",
+            "z" to "47"
+        )
+        assertEquals(res.metadata.evaluatedVariablesState, varsUpdate)
+        assertFalse(repl.notebook.variablesState.isEmpty())
+        val varsState = repl.notebook.variablesState
+        assertEquals("1", varsState["x"]!!.stringValue)
```
ileasile, line 467: Let's add extension method that will allow us to write `varsState.getStringValue("x")`. And also, `varsState.getValue("x")` for getting raw value

ileasile (commented):
Almost all is OK. The latest serious thing which I don't like is that stringified values are exposed as a part of Notebook API. They may be exposed, but at least with the corresponding raw values.

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/VariableState.kt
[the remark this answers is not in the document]
ileasile: Hm, I will keep this in mind, but we generally don't use it in this project

jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/compiler/util/serializedCompiledScript.kt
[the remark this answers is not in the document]
ileasile: Ok!

src/main/kotlin/org/jetbrains/kotlinx/jupyter/util.kt
nikolay-egorov: T and V is usually combo from Java, but ok

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
```diff
@@ -395,7 +415,14 @@ class ReplForJupyterImpl(
                 updateClasspath()
             } ?: emptyList()
 
-            EvalResult(rendered, EvaluatedSnippetMetadata(newClasspath, compiledData, newImports))
+            notebook.updateVariablesState(variablesState)
+            notebook.cellVariables = cellVariables
```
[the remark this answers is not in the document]
nikolay-egorov: It seems like we cant make it val since we need to reasign it once it changed

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
```diff
@@ -360,6 +361,22 @@ class ReplForJupyterImpl(
...
 
+    /**
+     * Used for debug purposes.
+     * @see ReplCommand
+     */
+    private fun printVariables(isHtmlFormat: Boolean = false) = log.debug(
+        if (isHtmlFormat) notebook.variablesReportAsHTML() else notebook.variablesReport()
+    )
+
+    // TODO: causes some tests to fail. perhaps, due to exhausting log calling
```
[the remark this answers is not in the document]
nikolay-egorov: Fixed

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
```diff
@@ -371,13 +393,16 @@ class ReplForJupyterImpl(
...
+                log.debug("Current cell id: $jupyterId")
+                executor.execute(code, displayHandler, currentCellId = jupyterId - 1) { internalId, codeToExecute ->
                     cell = notebook.addCell(internalId, codeToExecute, EvalData(jupyterId, code))
                 }
             } finally {
                 compiledData = internalEvaluator.popAddedCompiledScripts()
                 newImports = importsCollector.popAddedImports()
             }
+            val variablesState = internalEvaluator.variablesHolder
+            val cellVariables = internalEvaluator.cellVariables
```
ileasile, line 405: Is this variable needed?

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
```diff
@@ -371,13 +393,16 @@ class ReplForJupyterImpl(
...
-                executor.execute(code, displayHandler) { internalId, codeToExecute ->
+                log.debug("Current cell id: $jupyterId")
+                executor.execute(code, displayHandler, currentCellId = jupyterId - 1) { internalId, codeToExecute ->
                     cell = notebook.addCell(internalId, codeToExecute, EvalData(jupyterId, code))
                 }
             } finally {
                 compiledData = internalEvaluator.popAddedCompiledScripts()
                 newImports = importsCollector.popAddedImports()
             }
+            val variablesState = internalEvaluator.variablesHolder
```
ileasile, line 404: This variable may be inlined

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/Notebook.kt
```diff
@@ -11,6 +9,19 @@ interface Notebook {
...
+     */
+    val variablesState: Map<String, VariableState>
+
+    /**
+     * Stores info about useful variables in a cell.
+     * Key: cellId;
+     * Value: set of variable names.
+     * Useful <==> declarations + modifying references
+     */
+    var cellVariables: Map<Int, Set<String>>
```
ileasile, line 23: Do we still need it to be a `var`?

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
```diff
@@ -124,4 +140,50 @@ internal class InternalEvaluatorImpl(
...
+            val state = VariableStateImpl(property, cellClassInstance)
+            variablesWatcher.addDeclaration(cellId, property.name)
+
+            // it was val, now it's var
+            if (property is KMutableProperty1) {
+                variablesHolder.remove(property.name)
+            }
+
+            if (property !is KMutableProperty1) {
+                variablesHolder[property.name] = state
```
ileasile, line 175: `if`-`else`?

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/AbstractReplTest.kt
```diff
@@ -58,6 +59,14 @@ abstract class AbstractReplTest {
...
     }
 
+    protected fun Map<String, VariableState>.getStringValue(variableName: String): String? {
+        return get(variableName)?.stringValue
+    }
+
+    protected fun Map<String, VariableState>.getValue(variableName: String): Any? {
+        return get(variableName)?.value
+    }
+
```
ileasile, line 69: Let's replace these declarations to `testUtil.kt`, they may be useful in other tests as well

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/Notebook.kt
nikolay-egorov: We might create another field and make val in api. I think it might make sense since we do not want to allow change notebook.varsState from notebook

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/InternalEvaluatorImpl.kt
nikolay-egorov: Could you expand a bit more?

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/Notebook.kt
ileasile: Let's do so then

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
nikolay-egorov: Nah)

ileasile:
Thank you @nikolay-egorov! Merged manually as c6b1390e7c714b3f30312bb0647d20d074761654

PR #298, opened 2021-07-07
Conditional Gradle plugin tasks initialization

### LINKED ISSUE #255: Fix Gradle plugin for Multiplatform project
When applying Gradle plugin in MPP, following message appears:
```
Single JVM target not found in a multiplatform project
```
It's because multiplatform targets are not available right after applying Kotlin multiplatform plugin.
This may be fixed by running `addDependenciesIfNeeded()` after all configurations are done and targets are available.

ileasile:
Fixes #255 
@altavir could you please review and test it with your project(s)? I'll send you a plugin version to use.


altavir:
OK. I think I can build it myself since it seems like a regular jar.

I do not have problems in my projects even with the old plugin because I define kotlin platforms in the plugin application block (so it is before the jupyter plugin). The problem appeared when I tried to demonstrate a stand-alone plugin.

ileasile:
@altavir you may try version `0.10.0-94-1`

PR #303, opened 2021-07-13
Add separate repository for library descriptors

### LINKED ISSUE #254: Create a separate repository for descriptors
Currently, merging versions for libraries creates a lot of "noise" for this repository which makes it hard to track actual issues progress, also I believe that in the future, we will have a lot of libraries and it will be hard to track them without "bucket" mechanism like in SDKMan or scoop.

ileasile:
Fix #254


nikitinas (approved):
`kotlin-jupyter-libraries` repo should be moved under `Kotlin`

PR #313, opened 2021-07-20
Add support for IPC transport protocol

strangepleasures:


ileasile:
Could you please explain a bit how it can be used and where is it described?

strangepleasures:
> Could you please explain a bit how it can be used and where is it described?

Sorry, I've just realized that it's not ready yet. But the idea is to support both tcp and ipc transport protocols for communication between the kernel and the kernel manager. This can be specified in `jupyter_notebook_config.py` as follows:
```
c.KernelManager.transport = 'ipc'
```

altavir:
I am not sure it is a good idea to connect Kotlin and Python via IPC. They have separate memory management. Also, it won't do any good since IPC is not much faster than the TCP and we are talking about the system with such low message rates.

PR #314, opened 2021-07-22
Add Serializable Vars State with nested fields descriptors 

nikolay-egorov:


jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/compiler/util/serializedCompiledScript.kt
```diff
@@ -20,12 +20,22 @@ data class SerializedCompiledScriptsData(
     }
 }
 
+@Serializable
+data class SerializedVariablesState(
+    val name: String = "",
```
ileasile, line 25: Do we really need a name here if we have a name as a map key? Or maybe we don't need map key at all?

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
```diff
@@ -420,10 +424,11 @@ class ReplForJupyterImpl(
             notebook.updateVariablesState(internalEvaluator)
             // printVars()
             // printUsagesInfo(jupyterId, cellVariables[jupyterId - 1])
+            val entry = notebook.variablesState.entries.lastOrNull()
+            val serializedVarsState =  variablesSerializer.serializeVariableState(jupyterId - 1, entry?.key, entry?.value)
```
ileasile, line 428: Where is this variable used?

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
```diff
@@ -420,10 +424,11 @@ class ReplForJupyterImpl(
             notebook.updateVariablesState(internalEvaluator)
             // printVars()
             // printUsagesInfo(jupyterId, cellVariables[jupyterId - 1])
+            val entry = notebook.variablesState.entries.lastOrNull()
```
ileasile, line 427: It's not clear why it should be last, especially in unordered map

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/ApiTest.kt
```diff
@@ -59,7 +59,7 @@ class ApiTest : AbstractSingleReplTest() {
             "y" to "abc",
             "z" to "47"
         )
-        assertEquals(res.metadata.evaluatedVariablesState, varsUpdate)
+        assertEquals(res.metadata.evaluatedVariablesState.mapValues { it.value.value }, varsUpdate)
```
ileasile, line 62: I think it's worth making an extension method for this construction

src/main/kotlin/org/jetbrains/kotlinx/jupyter/serializationUtils.kt
```diff
@@ -0,0 +1,377 @@
...
+import kotlin.reflect.KTypeParameter
+import kotlin.reflect.full.declaredMemberProperties
+import kotlin.reflect.full.isSubclassOf
+import kotlin.reflect.jvm.isAccessible
+
+typealias FieldDescriptor = Map<String, SerializedVariablesState?>
+typealias MutableFieldDescriptor = MutableMap<String, SerializedVariablesState?>
+typealias PropertiesData = Collection<KProperty1<out Any, *>>
+
+data class ProcessedSerializedVarsState(
```
ileasile, line 18: Should it really be `data`? I mean, do you use it somewhere where `equals` and `hashCode` are needed, i.e. in map keys or in sets? If no, just make it a simple class and remove unnecessary overrides

ileasile (commented):
I didn't look through all the code, but I have some meta-comments.
Also, it seems that serialization may be performance-expensive, so let's introduce a flag (with the value from a system property and defaulting to true) that turns off the serialization (empty map should be sent in metadata)

nikolay-egorov:
> empty map should be sent in metadata
Maybe not empty, but just one-depth stringed values as it was before?

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
nikolay-egorov: Nothing, just for test)

src/main/kotlin/org/jetbrains/kotlinx/jupyter/serializationUtils.kt
nikolay-egorov: Fair enough

jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/compiler/util/serializedCompiledScript.kt
nikolay-egorov: I used key because it's really convenient to get the needed variable fast with only one string value. Perhaps, we might redux and remove the name

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
nikolay-egorov: Just for the test) just grab any variable

jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/compiler/util/serializedCompiledScript.kt
ileasile: Ok, let's leave it as is

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl.kt
ileasile: Let's then remove it, tests should go to tests sources :)

ileasile:
I think that empty is better. If the client does not support this functionality, it will not benefit from any data we'll send in this map.

jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/compiler/util/serializedCompiledScript.kt
nikolay-egorov: I think we still need to provide accessibility for names without a keymap if a user decides to use it for their own purposes. What do you think of leaving this extra info?

nikolay-egorov:
Wandering through the park, I came across this. We could completely go for jvm fields instead of KProperties since there are some cases when we need jvms

src/main/kotlin/org/jetbrains/kotlinx/jupyter/serializationUtils.kt
```diff
@@ -0,0 +1,406 @@
...
+        if (isFromJavaArray || kClass.isSubclassOf(Array::class)) {
+            value as Array<*>
+            return buildString {
+                val size = value.size
+                value.forEachIndexed { index, it ->
+                    print(this, size, index, it)
+                }
+            }
+        }
+        val isCollection = kClass.isSubclassOf(Collection::class)
```
ileasile, line 381: It is better to make extension properties on KClass and call them like this:
```
if (kClass.isCollection)
```

src/main/kotlin/org/jetbrains/kotlinx/jupyter/serializationUtils.kt
```diff
@@ -0,0 +1,406 @@
...
+            computedDescriptors[name] = createSerializeVariableState(name, null, callInstance)
+            descriptorStorage[name] = computedDescriptors[name]?.serializedVariablesState
+            true
+        } else {
+            false
+        }
+    }
+
+    companion object {
+        const val serializationEnvProperty = "jupyter.serialization.enabled"
```
ileasile, line 354: Not env, but system

src/main/kotlin/org/jetbrains/kotlinx/jupyter/serializationUtils.kt
```diff
@@ -0,0 +1,406 @@
...
+            property.get(callInstance)
+        } catch (e: Throwable) {
+            null
+        }
+        property.isAccessible = wasAccessible
+
+        return value
+    }
+
+    private fun checkCreateForPossibleArray(callInstance: Any, descriptorStorage: MutableFieldDescriptor, computedDescriptors: MutableMap<String, ProcessedSerializedVarsState>): Boolean {
```
ileasile, line 341: I don't like this naming, i.e. I don't understand `create` in the name

src/main/kotlin/org/jetbrains/kotlinx/jupyter/serializationUtils.kt
```diff
@@ -0,0 +1,406 @@
...
+            javaClass.declaredFields
+        } else { null }
+
+        val isContainer = if (membersProperties != null) (
+            membersProperties.isNotEmpty() || value!!::class.java.isArray || (javaClass != null && javaClass.isMemberClass)
+            ) else false
+        val type = if (value != null && value::class.java.isArray) {
+            "Array"
+        } else if (isContainer && value is List<*>) {
+            "SingletonList"
```
ileasile, line 309: Why is it "singleton" list?

ileasile (commented):
I have some more comments. Still not finished the review

src/main/kotlin/org/jetbrains/kotlinx/jupyter/util.kt
```diff
@@ -69,6 +70,10 @@ fun ResultsRenderersProcessor.registerDefaultRenderers() {
     register(bufferedImageRenderer)
 }
 
+fun Map<String, SerializedVariablesState>.getValuesToString(): Map<String, String?> {
```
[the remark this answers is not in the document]
nikolay-egorov: I like `mapValuesToStrings` a bit more

src/main/kotlin/org/jetbrains/kotlinx/jupyter/serializationUtils.kt
nikolay-egorov: Left from experiments. Bug, not a feature)

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/ReplVarsSerializationTest.kt
```diff
@@ -0,0 +1,494 @@
...
+
+    @Test
+    fun moreThanDefaultDepthContainerSerialization() {
+        val res = eval(
+            """
+            val x = listOf(listOf(1), listOf(2), listOf(3), listOf(4))
+            """.trimIndent(),
+            jupyterId = 1
+        )
+        val varsData = res.metadata.evaluatedVariablesState
```
ileasile, line 79: Generally, tests are too long. I'm not sure that each test should test all internals. Each test should check one or two things

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/ReplVarsSerializationTest.kt
```diff
@@ -0,0 +1,494 @@
...
+            state.isContainer shouldBe true
+            state.value shouldBe "${values++}"
+        }
+
+        val depthMostNode = actualContainer.fieldDescriptor.entries.first { it.value!!.isContainer }
+        val serializationAns = serializer.doIncrementalSerialization(0, "x", depthMostNode.key, depthMostNode.value!!)
+    }
+
+    @Test
+    fun incrementalUpdateTestWithPath() {
```
ileasile, line 167: This test seems mostly the same as the previous...

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/ReplTests.kt
```diff
@@ -415,29 +416,9 @@ class ReplTests : AbstractSingleReplTest() {
         (res as (Int) -> Int)(1) shouldBe 2
     }
 
-    @Test
```
ileasile, line 418: Why did you remove these tests? Oo

src/main/kotlin/org/jetbrains/kotlinx/jupyter/protocol.kt
```diff
@@ -513,3 +547,8 @@ fun JupyterConnection.evalWithIO(repl: ReplForJupyter, srcMessage: Message, body
         KernelStreams.setStreams(false, out, err)
     }
 }
+
+fun EvaluatedSnippetMetadata?.convertToNullIfEmpty(): JsonElement? {
+    val jsonNode = Json.encodeToJsonElement(this)
+    return if (jsonNode is JsonNull || jsonNode?.jsonObject.isEmpty()) null else jsonNode
```
ileasile, line 553:
```kotlin
return Json.encodeToJsonElement(this).takeIf { it !is JsonNull && it.jsonObject.isNotEmpty() }
```

src/main/kotlin/org/jetbrains/kotlinx/jupyter/protocol.kt
```diff
@@ -321,6 +344,17 @@ fun JupyterConnection.Socket.shellMessagesHandler(msg: Message, repl: ReplForJup
                 }
             }
         }
+        is SerializationRequest -> {
```
ileasile, line 347: Are SerializationRequest's are still used? I thought we switched to comm-s

src/main/kotlin/org/jetbrains/kotlinx/jupyter/apiImpl.kt
```diff
@@ -149,6 +152,11 @@ class NotebookImpl(
     override val jreInfo: JREInfoProvider
         get() = JavaRuntime
 
+    fun updateVariablesState(evaluator: InternalEvaluator) {
```
ileasile, line 155: It seems like a very internal method. We most likely don't want to expose it to user

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/ReplVarsTest.kt
```diff
@@ -318,4 +321,121 @@ class ReplVarsTest : AbstractSingleReplTest() {
...
+        state.size shouldBe 0
+
+        eval(
+            """
+            var f = 47 
+            """.trimIndent(),
+            jupyterId = 3
+        )
+        state = repl.notebook.unchangedVariables
+        // tmp disable to further investigation (locally tests pass on java8)
```
ileasile, line 438: It also passes for me locally. I think ignoring is a bad solution

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/ReplVarsTest.kt
```diff
@@ -318,4 +321,121 @@ class ReplVarsTest : AbstractSingleReplTest() {
...
+        eval(
+            """
+            private val x = "abcd"
+            var f = 47
+            internal val z = 47
+            """.trimIndent(),
+            jupyterId = 1
+        )
+        var state = repl.notebook.unchangedVariables
+        state.size.shouldBe(3)
```
ileasile, line 418: Actually, what do you test here? What variables should be contained in this list? What if this test fail, how will you understand what variables have changed without debugging?

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/ReplVarsTest.kt
```diff
@@ -318,4 +321,121 @@ class ReplVarsTest : AbstractSingleReplTest() {
...
+    fun unchangedVariablesGapedRedefinition() {
+        eval(
+            """
+            private val x = "abcd"
+            var f = 47
+            internal val z = 47
+            """.trimIndent(),
+            jupyterId = 1
+        )
+        var state = repl.notebook.unchangedVariables
```
ileasile, line 417: Also, why do you need this variable? It's useless

src/test/kotlin/org/jetbrains/kotlinx/jupyter/test/repl/ReplVarsTest.kt
```diff
@@ -318,4 +321,121 @@ class ReplVarsTest : AbstractSingleReplTest() {
         varState.getStringValue("x") shouldBe "25.0"
         varState.getValue("x") shouldBe 25.0
     }
+
+    @Test
+    fun testAnonymousObjectRendering() {
```
ileasile, line 326: This is the test from other file that was deleted

ileasile (changes requested):
Please be more attentive when fix issues. Most of them are global, and should be fixed not only in one place but all across the code

PR #329, opened 2021-09-06
Fix #326 - proper handling of store_history

### LINKED ISSUE #326: Every execution request increments `execution_count`,  even if `store_history=False`
According to Jupyter documentation, `execution_count` should be incremented only when `store_history=True`:
"The kernel should have a single, monotonically increasing counter of all execution requests that are made with store_history=True. This counter is used to populate the In[n] and Out[n] prompts. The value of this counter will be returned as the execution_count field of all execute_reply and execute_input messages.".

However, currently even requests with `store_history=False` increment the counter.

strangepleasures:


ileasile:
Hi, Pavel! How do you use `store_history`? Are you interested only in not incrementing the counter, or do you really wish not to save compiled snippets in memory?

strangepleasures:
> Hi, Pavel! How do you use `store_history`? Are you interested only in not incrementing the counter, or do you really wish not to save compiled snippets in memory?

Hi Ilya, in principle, I only need proper numbering (consistent with `Out[]` values of course). If you need to keep the outputs for any other means, e.g. rendering, I guess it could be a separate data structure.

ileasile:
@strangepleasures I've changed the logic a bit and slightly refactored the code. Could you please review new logic? It's in this test:
https://github.com/Kotlin/kotlin-jupyter/pull/329/files#diff-374830c795b7e656d78a644f288ac3108fdb5048764f61ee5b3d2a8ddd54e973R314-R330

strangepleasures (commented):
Looks good!

ileasile:
Well, I'm merging it then

PR #334, opened 2021-09-22
Fix #333 Close HTTP connections

### LINKED ISSUE #333: Leaking HTTP connections
There's a lot of places in the code where HTTP(S) requests are made using `getHttp`. 
Contrarily to the common best practice, HTTP connections are never closed explicitly.
They *might* get closed by the system at some point, but that somehow never happens in our AWS setup. 
(May be OS/JVM/settings-dependent).

strangepleasures:


build-plugin/common-dependencies/src/org/jetbrains/kotlinx/jupyter/common/httpUtil.kt
```diff
@@ -22,18 +22,18 @@ class ResponseWrapper(
...
     return ClientFilters.FollowRedirects().then(ApacheClient())
 }
 
-fun httpRequest(request: Request): ResponseWrapper {
+fun <R> httpRequest(request: Request, mapper: (ResponseWrapper) -> R): R {
     val client = createHttpClient()
     val response = client(request)
 
-    return ResponseWrapper(response, request.uri.toString())
+    return ResponseWrapper(response, request.uri.toString()).use(mapper)
```
ileasile, line 33: Won't it fix the issue if we just close the response here? I mean
```
val client = createHttpClient()
val response = client(request)
response.close()
return ResponseWrapper(response, request.uri.toString())
```

strangepleasures: The connection will be closed of course, but I'm afraid that if you try to access the request's body after that, you'll get an exception

strangepleasures: Actually, it seems that closing the response doesn't help at this point. The returned response object is a MemoryResponse, so closing it is a no-op. The HTTP connection is probably expected to be closed in `CloseableHttpResponse.toHttp4kResponse` which consumes the entire payload, but again, that doesn't happen in our AWS setup.

ileasile: @strangepleasures have you checked that the edits you have introduced fix the issue in your AWS setup?

ileasile: I can provide you with pip package if you need

strangepleasures: `ApacheClient()` without arguments creates a client with a `PoolingHttpClientConnectionManager` with infinite TTL. Maybe, that's why connections never get closed.

PR #337, opened 2021-09-29
Add exceptions rendering

### LINKED ISSUE #332: Add exception handling/interception
Currently there is no easy way to format how exceptions that are thrown by code are displayed in the output of the cell.

It would be good from a usability point of view to have some influence on this, since the amount of info that is typically shown (like including full stack trace) doesn't often make much sense in the context of a notebook. One example is require (IllegalArgument exceptions). These could be presented in a much nicer way with focus on the message and perhaps no stack trace at all or collapsed. 

So request to add an Exception API (for example to the JupyterIntegration class) that makes possible to handle exceptions and transform them into something else.

See also a short discussion on: https://kotlinlang.slack.com/archives/C4W52CFEZ/p1631949623037000

ileasile:
This PR adds the ability to render exceptions via API. Renderers should return `DisplayResult`

Fix #332


ileasile:
image

PR #364, opened 2022-05-02
integrationTypeNameRules support for LibrariesScanner

Jolanrensen:
Adds support for integrationTypeNameRules inside JUnit tests. Might need some extra work, but it's a start :)


Jolanrensen:
I see that using just this, the other library instances also get initiated. They don't get executed I think (my tests work fine), so probably I missed something.

Jolanrensen:
It works great! IMO you can merge it

PR #375, opened 2022-05-28
Add comm-message handlers and support for debugPort config retrieval

nikolay-egorov:

PR #376, opened 2022-06-20
Add socket messages callbacks and comms API

### LINKED ISSUE #366: Support Comm messages in the Kernel API
Support [Comm messages](https://jupyter-notebook.readthedocs.io/en/stable/comms.html) in the Kernel API. They will help to make interactive dataframes in the Kernel

ileasile:
Fixes #366

Should also partially fix the needs of #375

### DIFF
--- a/jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/libraries/connection.kt
+++ b/jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/libraries/connection.kt
@@ -1,0 +1,192 @@
+package org.jetbrains.kotlinx.jupyter.api.libraries
+
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.decodeFromJsonElement
+import kotlinx.serialization.json.encodeToJsonElement
+import kotlinx.serialization.json.jsonObject
+import org.jetbrains.kotlinx.jupyter.util.EMPTY
+
+/**
+ * Jupyter connection socket types
+ * Here you can find an information about Jupyter sockets:
+ * https://jupyter-client.readthedocs.io/en/stable/messaging.html#introduction
+ *
+ * For now, only adding callbacks for messages on `control` and `shell` sockets makes sense.
+ */
+enum class JupyterSocket {
+    HB,
+    SHELL,
+    CONTROL,
+    STDIN,
+    IOPUB;
+}
+
+/**
+ * Raw Jupyter message. [data] generally should contain `header`, `parent_header`, `content` and `metadata` fields
+ *
+ * @constructor Create empty Raw message
+ */
+interface RawMessage {
+    val id: List<ByteArray>
+    val data: JsonElement
+}
+
+val RawMessage.header: JsonObject?
+    get() = (data as? JsonObject)?.get("header") as? JsonObject
+
+val RawMessage.type: String?
+    get() {
+        val type = header?.get("msg_type")
+        if (type !is JsonPrimitive || !type.isString) return null
+        return type.content
+    }
+
+val RawMessage.content: JsonObject?
+    get() = (data as? JsonObject)?.get("content") as? JsonObject
+
+typealias CommOpenCallback = (Comm, JsonObject) -> Unit
+typealias CommMsgCallback = (JsonObject) -> Unit
+typealias CommCloseCallback = (JsonObject) -> Unit
+typealias RawMessageAction = (RawMessage) -> Unit
+
+/**
+ * Callback for messages of type [messageType] coming to a certain [socket]
+ */
+interface RawMessageCallback {
+    val socket: JupyterSocket
+    val messageType: String
+    val action: RawMessageAction
+}
+
+interface JupyterConnection {
+    /**
+     * Add callback for incoming message and return it
+     */
+    fun addMessageCallback(callback: RawMessageCallback): RawMessageCallback
+
+    /**
+     * Remove added message callback
+     */
+    fun removeMessageCallback(callback: RawMessageCallback)
+
+    /**
+     * Send raw [message] to a given [socketName]
+     */
+    fun send(socketName: JupyterSocket, message: RawMessage)
+
+    /**
+     * Send reply to a given [parentMessage] of type [type] to socket [socketName].
+     * Simpler-to-use version of [send].
+     */
+    fun sendReply(socketName: JupyterSocket, parentMessage: RawMessage, type: String, content: JsonObject, metadata: JsonObject? = null)
+}
+
+interface CommManager {
+    /**
+     * Creates a comm with a given target, generates unique ID for it. Sends comm_open request to frontend
+     *
+     * @param target Target to create comm for. Should be registered on frontend side.
+     * @param data Content of comm_open message
+     * @return Created comm
+     */
+    fun openComm(target: String, data: JsonObject = Json.EMPTY): Comm
+
+    /**
+     * Closes a comm with a given ID. Sends comm_close request to frontend
+     *
+     * @param id ID of a comm to close
+     * @param data Content of comm_close message
+     */
+    fun closeComm(id: String, data: JsonObject = Json.EMPTY)
+
+    /**
+     * Get all comms for a given target, or all opened comms if `target` is `null`
+     */
+    fun getComms(target: String? = null): Collection<Comm>
+
+    /**
+     * Register a [callback] for `comm_open` with a specified [target]. Overrides already registered callback.
+     *
+     * @param target
+     * @param callback
+     */
+    fun registerCommTarget(target: String, callback: CommOpenCallback)
+
+    /**
+     * Unregister target callback
+     */
+    fun unregisterCommTarget(target: String)
+}
+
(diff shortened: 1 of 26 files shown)


jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/libraries/connection.kt
```diff
@@ -0,0 +1,117 @@
...
+    val target: String
+    val id: String
+    fun sendJson(data: JsonObject)
+    fun onMessage(action: CommMsgCallback)
+    fun onClose(action: CommCloseCallback)
+
+    // Closes a comm. Sends comm_close request to frontend
+    fun close(data: JsonObject = emptyJsonObject, notifyClient: Boolean = true)
+
+    fun messageReceived(data: JsonObject)
```
altavir, line 101: What does it do? It is not obvious.

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/libraries/connection.kt
```diff
@@ -0,0 +1,117 @@
...
+     */
+    fun getComms(target: String? = null): Collection<Comm>
+    fun registerTarget(target: String, callback: CommOpenCallback)
+}
+
+interface Comm {
+    val target: String
+    val id: String
+    fun sendJson(data: JsonObject)
+    fun onMessage(action: CommMsgCallback)
```
altavir, line 95: Probably need to clarify if new `onMessage` erases the old one or is added to it.

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/util/serializers.kt
```diff
@@ -21,6 +21,8 @@ import org.jetbrains.kotlinx.jupyter.api.libraries.ResourceFallbacksBundle
 import kotlin.reflect.KClass
 import kotlin.reflect.KProperty1
 
+val emptyJsonObject = JsonObject(mapOf())
```
altavir, line 24: It pollutes the global namespace. Probably better to make it internal or hide behind `Json.Companion.EMPTY`

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/libraries/connection.kt
```diff
@@ -0,0 +1,117 @@
...
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.decodeFromJsonElement
+import kotlinx.serialization.json.encodeToJsonElement
+import kotlinx.serialization.json.jsonObject
+import org.jetbrains.kotlinx.jupyter.util.emptyJsonObject
+
+enum class JupyterSocket {
```
altavir, line 12: What different connection types do?

altavir (commented):
There should be some kind of documentation about how those messages are used from the frontend part of the notebook.

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/libraries/connection.kt
ileasile: I've removed it from public API

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/util/serializers.kt
ileasile: Added `Json.EMPTY` extension intead

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/libraries/connection.kt
```diff
@@ -0,0 +1,190 @@
...
+import org.jetbrains.kotlinx.jupyter.util.EMPTY
+
+/**
+ * Jupyter connection socket types
+ * Here you can find an information about Jupyter sockets:
+ * https://jupyter-client.readthedocs.io/en/stable/messaging.html#introduction
+ *
+ * For now, only adding callbacks for messages on `control` and `shell` sockets makes sense.
+ */
+enum class JupyterSocket {
```
nikolay-egorov, line 19: Looks like this enum provides same info as `JupyterSocketInfo`, perhaps, it's better to incorporate one into another

src/main/kotlin/org/jetbrains/kotlinx/jupyter/connection.kt
```diff
@@ -150,14 +227,208 @@ class JupyterConnection(val config: KernelConfig) : Closeable {
...
+                )
+            )
+        )
+        send(socketName, message)
+    }
+
+    private val commOpenCallbacks = ConcurrentHashMap<String, CommOpenCallback>()
+    private val commTargetToIds = ConcurrentHashMap<String, CopyOnWriteArrayList<String>>()
+    private val commIdToComm = ConcurrentHashMap<String, CommImpl>()
+    override fun openComm(target: String, data: JsonObject): Comm {
```
nikolay-egorov, line 369: New line missing

src/main/kotlin/org/jetbrains/kotlinx/jupyter/connection.kt
```diff
@@ -150,14 +227,208 @@ class JupyterConnection(val config: KernelConfig) : Closeable {
...
+            makeSimpleMessage(
+                MessageType.COMM_OPEN,
+                CommOpen(newComm.id, newComm.target)
+            )
+        )
+
+        return newComm
+    }
+
+    fun processCommOpen(target: String, id: String, data: JsonObject): Comm {
```
nikolay-egorov, line 384: Here we assume that registration is always successful, but if there is an exception, we do not do anything. According to [doc](https://jupyter-client.readthedocs.io/en/latest/messaging.html#opening-a-comm), if it fails we actually might send a commClose to notify failure

src/main/kotlin/org/jetbrains/kotlinx/jupyter/connection.kt
```diff
@@ -150,14 +227,208 @@ class JupyterConnection(val config: KernelConfig) : Closeable {
...
+
+    override fun getComms(target: String?): Collection<Comm> {
+        return if (target == null) {
+            commIdToComm.values.toList()
+        } else {
+            commTargetToIds[target].orEmpty().mapNotNull { commIdToComm[it] }
+        }
+    }
+
+    fun processCommMessage(id: String, data: JsonObject) {
```
nikolay-egorov, line 421: There might be some heavy computations, including execution or interpreter thread work, so maybe it's good to add a possibility to indicate such cases in a signature and send `STATUS.BUSY `

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/libraries/connection.kt
ileasile: `JupyterSocketInfo` has some implementation details which I don't want to expose to the public API. I've incorporated one into another however

jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/util/serializers.kt
```diff
@@ -21,6 +21,11 @@ import org.jetbrains.kotlinx.jupyter.api.libraries.ResourceFallbacksBundle
 import kotlin.reflect.KClass
 import kotlin.reflect.KProperty1
 
+private val emptyJsonObject = JsonObject(mapOf())
+
+@Suppress("unused")
+val Json.EMPTY get() = emptyJsonObject
```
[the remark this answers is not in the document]
ileasile: It will be present in *-api artifact, so it'll be available in IDEA plugin, inside notebooks and for library writers.

src/main/kotlin/org/jetbrains/kotlinx/jupyter/execution.kt
```diff
@@ -0,0 +1,72 @@
...
+}
+
+class JupyterExecutorImpl : JupyterExecutor {
+    private val currentExecutions: MutableSet<Thread> = Collections.newSetFromMap(ConcurrentHashMap())
+    private val coroutineScope = CoroutineScope(Dispatchers.Default)
+
+    override fun <T> runExecution(classLoader: ClassLoader?, body: () -> T): ExecutionResult<T> {
+        var execRes: T? = null
+        var execException: Throwable? = null
+        val execThread = thread(contextClassLoader = classLoader ?: Thread.currentThread().contextClassLoader) {
```
nikolay-egorov, line 31: Sorry for posting a bit late. This would work, but creating and stopping a new thread a bit costly since it's OS call. 

The more consise would be to reuse existing threads, so they wait for some tasks to appear from a queue and execute them, e.g. pool aproach. So perhaps we might reuse some existing, like `Executors.newCachedThreadPool `

ileasile: Yes, I undestand. It's just how it was implemented before. I will change it in a separate commit

PR #382, opened 2022-08-17
Add Gradle-like API for adding dependencies

### LINKED ISSUE #367: Add a function to configure dependencies and repositories in gradle-like manner
File-level annotations seem to be rather artificial for the notebook. I propose to add a configuration function that will contatin all necesary repositories and dependencies notations. So instead of writing 

```kotlin
@file:Repository("https://repo.kotlin.link")
@file:DependsOn("space.kscience:plotlykt-jupyter:0.5.0")
```
we could do 

```kotlin
notebook.configure{
  repositories{
    maven("https://repo.kotlin.link")
  }
  dependencies{
    implementation("space.kscience:plotlykt-jupyter:0.5.0")
  }
}
```
I specifically made the design exatly the same as gradle configuration. People could just copy configuration blocks from library descriptions or even from `mavenCentral`.

The configuration could be simplified with some shortcuts 

```kotlin
notebook.repositories.maven("https://repo.kotlin.link")
notebook.dependencies += "space.kscience:plotlykt-jupyter:0.5.0"
```

Later some additional configuraion options could be added in the same configuration block.

ileasile:
Fix #367

### DIFF
--- a/jupyter-lib/lib/src/main/kotlin/jupyter/kotlin/ScriptTemplateWithDisplayHelpers.kt
+++ b/jupyter-lib/lib/src/main/kotlin/jupyter/kotlin/ScriptTemplateWithDisplayHelpers.kt
@@ -20,7 +20,7 @@ abstract class ScriptTemplateWithDisplayHelpers(
 
     fun EXECUTE(code: String) = host.scheduleExecution(CodeExecution(code).toExecutionCallback())
 
-    fun USE(library: LibraryDefinition) = host.addLibrary(library)
+    fun USE(library: LibraryDefinition) = host.scheduleExecution { addLibrary(library) }
 
     fun USE(builder: JupyterIntegration.Builder.() -> Unit) {
         val o = object : JupyterIntegration() {
--- a/jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/libraries/gradleLike.kt
+++ b/jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/libraries/gradleLike.kt
@@ -1,0 +1,45 @@
+package org.jetbrains.kotlinx.jupyter.api.libraries
+
+import java.io.File
+
+interface RepositoryHandlerScope {
+    fun maven(url: String)
+    fun dir(dir: File)
+}
+
+interface DependencyHandlerScope {
+    fun implementation(coordinates: String)
+    fun implementation(group: String, artifact: String, version: String)
+}
+
+fun RepositoryHandlerScope.mavenCentral() = maven("https://repo.maven.apache.org/maven2/")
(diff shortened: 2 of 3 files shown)


jupyter-lib/lib/src/main/kotlin/jupyter/kotlin/ScriptTemplateWithDisplayHelpers.kt
```diff
@@ -20,7 +20,7 @@ abstract class ScriptTemplateWithDisplayHelpers(
 
     fun EXECUTE(code: String) = host.scheduleExecution(CodeExecution(code).toExecutionCallback())
 
-    fun USE(library: LibraryDefinition) = host.addLibrary(library)
+    fun USE(library: LibraryDefinition) = host.scheduleExecution { addLibrary(library) }
```
ileasile, line 23: If we try to add library immediately, recursive execution happens - execution DURING another execution. We don't know how to handle it, because we use previous script instances for evaluation, and it's not clear what to do without them. So we just schedule the execution to the end of the current execution.

nikolay-egorov (approved):
Looks neat!

ileasile:
I'm concerned about names too: it's not clear what `gradle` means, it sounds like we configure Gradle, not kernel. But I can't come up with a better name for this DSL entry point
Maven resolver is used underhood, yes.

altavir:
> I'm concerned about names too: it's not clear what `gradle` means, it sounds like we configure Gradle, not kernel. But I can't come up with a better name for this DSL entry point Maven resolver is used underhood, yes.

maybe something like `requires`? It does not matter how it is called since you copy only what is inside.

ileasile:
1. `gradle` - clear what you should write inside but doesn't explain the purpose
2. `gradleLike[Conf]` - better, but verbose
3. `requires` - good borrowing from JS modules, but there is `require` in Kotlin, and may be misleading
4. `config`, `configure`, `provide` - not bad, but too generic
5. `build` - looks like a compromise variant
6. `dependencyResolutionManagement` - name of block in `settings.gradle[.kts]` that is actually responsible for dependency management

Let's choose something please

altavir:
@ileasile I think it is better to discuss in slack. Maybe a vote. 

I do not think that the outer scope should match how it looks in Gradle because people do not copy the whole block, they copy its contents. Also `dependencyResolutionManagement` is not used in project config, only in `settings`. 

What about `dependsOn{ }`?

ileasile:
Final version of API:
```kotlin
USE {
	repositories {
		mavenCentral()
	}
	dependencies {
		implementation("io.github.config4k:config4k:0.4.2")
	}
}
```

Or in integration:
```kotlin
internal class Integration : JupyterIntegration() {
    
    override fun Builder.onLoaded() {
        repositories {
		mavenCentral()
	}
	dependencies {
		implementation("io.github.config4k:config4k:0.4.2")
	}
    }
}
```

PR #404, opened 2023-03-14
Moved cache in message_type serializer to ConcurrentHashMap

jbaron:
Moved cache in message_type serializer to ConcurrentHashMap to avoid concurrent modification exceptions. ConcurrentHashMap seems a good fit due to low overhead when reading from the map (which is the main purpose of a cache).

Updated both MessageTypeSerializer and DetailsLevelSerializer.


ileasile:
Thank you for PR! Please run Reformat.. run configuration, I think there's an issue with imports order. Also, run :check instead of :test, it also checks formatting and Readme consistency. And it's actually run on TC

ileasile:
Yes, it's some order that our formatter respects :)

PR #420, opened 2023-06-29
Skip HTML files when processing this project in linguist

DRSchlaubi:
Currently according to linguist this project is written in 98% HTML, because of 2 large HTML files which are probabbly generated


ileasile:
Thanks!

PR #447, opened 2024-02-06
Introduce class for setting up debug integration

nikolay-egorov:
This way, newer version of kernel will keep thread for creating breakpoints

### DIFF
--- a/jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/notebook/impl/NotebookImpl.kt
+++ b/jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/notebook/impl/NotebookImpl.kt
@@ -24,6 +24,7 @@ import org.jetbrains.kotlinx.jupyter.api.libraries.Variable
 import org.jetbrains.kotlinx.jupyter.codegen.FieldsProcessorInternal
 import org.jetbrains.kotlinx.jupyter.codegen.ResultsRenderersProcessor
 import org.jetbrains.kotlinx.jupyter.codegen.TextRenderersProcessorWithPreventingRecursion
+import org.jetbrains.kotlinx.jupyter.debug.DEBUG_THREAD_NAME
 import org.jetbrains.kotlinx.jupyter.libraries.parseLibraryDescriptor
 import org.jetbrains.kotlinx.jupyter.repl.EvalData
 import org.jetbrains.kotlinx.jupyter.repl.ReplRuntimeProperties
@@ -31,13 +32,30 @@ import org.jetbrains.kotlinx.jupyter.repl.SharedReplContext
 import org.jetbrains.kotlinx.jupyter.repl.notebook.MutableCodeCell
 import org.jetbrains.kotlinx.jupyter.repl.notebook.MutableNotebook
 import org.jetbrains.kotlinx.jupyter.util.JupyterClientDetector
+import java.io.Closeable
+import kotlin.concurrent.thread
 
 class NotebookImpl(
     private val runtimeProperties: ReplRuntimeProperties,
     override val commManager: CommManager,
     private val explicitClientType: JupyterClientType?,
     override val libraryLoader: LibraryLoader,
-) : MutableNotebook {
+    debugPortProvided: Boolean,
+) : MutableNotebook, Closeable {
+    private fun `$debugMethod`() {
+        try {
+            while (true) {
+                Thread.sleep(1000)
+            }
+        } catch (_: InterruptedException) {
+        }
+    }
+    private val debugThread = if (debugPortProvided) {
+        thread(
+            name = DEBUG_THREAD_NAME,
+            block = ::`$debugMethod`,
+        )
+    } else null
     private val cells = hashMapOf<Int, MutableCodeCell>()
     override var sharedReplContext: SharedReplContext? = null
 
@@ -169,5 +187,9 @@ class NotebookImpl(
     override fun getLibraryFromDescriptor(descriptorText: String, options: Map<String, String>): LibraryDefinition {
         return parseLibraryDescriptor(descriptorText)
             .convertToDefinition(options.entries.map { Variable(it.key, it.value) })
+    }
+
+    override fun close() {
+        debugThread?.interrupt()
     }
 }
--- a/src/main/kotlin/org/jetbrains/kotlinx/jupyter/messaging/MessageHandlerImpl.kt
+++ b/src/main/kotlin/org/jetbrains/kotlinx/jupyter/messaging/MessageHandlerImpl.kt
@@ -1,9 +1,11 @@
 package org.jetbrains.kotlinx.jupyter.messaging
 
 import org.jetbrains.kotlinx.jupyter.api.libraries.RawMessage
+import org.jetbrains.kotlinx.jupyter.closeIfPossible
 import org.jetbrains.kotlinx.jupyter.execution.JupyterExecutor
 import org.jetbrains.kotlinx.jupyter.messaging.comms.CommManagerInternal
 import org.jetbrains.kotlinx.jupyter.repl.ReplForJupyter
+import java.io.Closeable
 import java.util.concurrent.atomic.AtomicLong
 
 class MessageHandlerImpl(
@@ -12,7 +14,7 @@ class MessageHandlerImpl(
     private val messageFactoryProvider: MessageFactoryProvider,
     private val socketManager: JupyterBaseSockets,
     private val executor: JupyterExecutor,
-) : AbstractMessageHandler() {
+) : AbstractMessageHandler(), Closeable {
     private val executionCount = AtomicLong(1)
 
     override fun createProcessor(message: RawMessage): MessageRequestProcessor {
@@ -25,5 +27,9 @@ class MessageHandlerImpl(
             executionCount,
             repl,
         )
+    }
+
+    override fun close() {
+        repl.closeIfPossible()
     }
 }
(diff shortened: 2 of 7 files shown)


jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/creating/LazilyConstructibleReplComponentsProvider.kt
```diff
@@ -29,8 +30,10 @@ abstract class LazilyConstructibleReplComponentsProvider : ReplComponentsProvide
     override val homeDir: File? by lazy { provideHomeDir() }
     protected abstract fun provideHomeDir(): File?
 
-    override val debugPort: Int? by lazy { provideDebugPort() }
-    protected abstract fun provideDebugPort(): Int?
+    override val debugPort: Int? by lazy { debugIntegrationProvider.provideDebugPort() }
```
ileasile, line 33: No, that's not how this pattern works. You can remove `val debugPort` completely if you wish. But just observe the code: is there any lazy implementation other than `lazy { provide*() }`?

jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/debug/DebugIntegrationProviderMock.kt
```diff
@@ -0,0 +1,3 @@
+package org.jetbrains.kotlinx.jupyter.debug
+
+object DebugIntegrationProviderMock : DebugIntegrationProvider(null)
```
ileasile, line 3: Looks more like a singleton property bc you actually don't override anything

jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/debug/DebugIntegration.kt
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    fun provideDebugPort(): Int? = debugPort
+
+    fun provideSuspendedThread(runnable: () -> Unit): Thread {
+        return thread(start = provideDebugPort() != null, name = DEBUG_THREAD_NAME, block = runnable)
+    }
+}
+
+class NotebookDebugIntegrationProvider(
```
ileasile, line 19: If it's a `NotebookDebugIntegrationProvider` then why `DebugIntegrationProvider` is missing "Notebook" part in its name?

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/creating/creatingUtil.kt
```diff
@@ -37,7 +39,9 @@ fun createRepl(
         override fun provideIsEmbedded() = isEmbedded
         override fun provideDisplayHandler() = displayHandler
         override fun provideCommunicationFacility(): JupyterCommunicationFacility = communicationFacility
-        override fun provideDebugPort(): Int? = debugPort
+        override fun provideDebugIntegration(): DebugIntegrationProvider {
+            return NotebookDebugIntegrationProvider(debugPort)
```
ileasile, line 43: Once again, is there anything around this code that doesn't simply use passed constructor parameter?

jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/debug/DebugIntegration.kt
```diff
@@ -0,0 +1,21 @@
...
+import kotlin.concurrent.thread
+
+abstract class DebugIntegrationProvider(
+    protected val debugPort: Int?,
+) {
+    companion object {
+        const val DEBUG_THREAD_NAME = "NotebookDebugThread"
+    }
+
+    fun provideDebugPort(): Int? = debugPort
```
ileasile, line 12: It looks like debugPort doesn't need wrappers like this

ileasile: Or maybe we don't need to provide debug port here at all? It looks like the logic is boolean here: if we want or don't want to create a debug thread

jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/debug/DebugIntegration.kt
```diff
@@ -0,0 +1,21 @@
...
+
+    fun provideSuspendedThread(runnable: () -> Unit): Thread {
```
ileasile, line 14: This abstraction looks leaking. I would suggest to have just a boolean flag here and create the thread in place it's needed

ileasile (changes requested):
For now it looks like we don't need debugPort in options at all. We need boolean instead. Abstraction DebugIntegrationProvider seems not needed

jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/creating/LazilyConstructibleReplComponentsProvider.kt
nikolay-egorov: I agree, that's a violation of the pattern. Probably, we might need debug port for future cases in embedded kernel

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/creating/creatingUtil.kt
nikolay-egorov: True, can be done. I'll change. However, I didn't think about this as a pattern to follow

nikolay-egorov:
I'll think about it one more time. Now it indeed seems like we don't need abstraction for debug just yet

jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/messaging/MessageHandler.kt
```diff
@@ -2,7 +2,8 @@ package org.jetbrains.kotlinx.jupyter.messaging
 
 import org.jetbrains.kotlinx.jupyter.api.libraries.JupyterSocketType
 import org.jetbrains.kotlinx.jupyter.api.libraries.RawMessage
+import java.io.Closeable
 
-interface MessageHandler {
+interface MessageHandler : Closeable {
```
ileasile, line 7: Interfaces shouldn't be closeable, implementations should be

jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/notebook/MutableNotebook.kt
```diff
@@ -6,8 +6,9 @@ import org.jetbrains.kotlinx.jupyter.codegen.ResultsRenderersProcessor
 import org.jetbrains.kotlinx.jupyter.codegen.TextRenderersProcessorWithPreventingRecursion
 import org.jetbrains.kotlinx.jupyter.repl.EvalData
 import org.jetbrains.kotlinx.jupyter.repl.SharedReplContext
+import java.io.Closeable
 
-interface MutableNotebook : Notebook {
+interface MutableNotebook : Notebook, Closeable {
```
ileasile, line 11: And for this interface as well

jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/notebook/impl/NotebookImpl.kt
```diff
@@ -24,20 +24,36 @@ import org.jetbrains.kotlinx.jupyter.api.libraries.Variable
...
 ) : MutableNotebook {
+    private fun `$debugMethod`() {
+        try {
+            while (true) {
+                Thread.sleep(1000)
+            }
+        } catch (_: InterruptedException) {
+        }
+    }
+    private val debugThread = thread(
```
ileasile, line 52: Why should we create this object if we even don't run it? I suggest to make it nullable

src/main/kotlin/org/jetbrains/kotlinx/jupyter/messaging/MessageHandlerImpl.kt
```diff
@@ -3,11 +3,11 @@ package org.jetbrains.kotlinx.jupyter.messaging
 import org.jetbrains.kotlinx.jupyter.api.libraries.RawMessage
 import org.jetbrains.kotlinx.jupyter.execution.JupyterExecutor
 import org.jetbrains.kotlinx.jupyter.messaging.comms.CommManagerInternal
-import org.jetbrains.kotlinx.jupyter.repl.ReplForJupyter
+import org.jetbrains.kotlinx.jupyter.repl.impl.ReplForJupyterImpl
 import java.util.concurrent.atomic.AtomicLong
 
 class MessageHandlerImpl(
-    private val repl: ReplForJupyter,
+    private val repl: ReplForJupyterImpl,
```
ileasile, line 10: Don't do this, it's a bad path. Just add an extension
```
fun Any.closeIfNeeded() {
    if (this is Closeable) close()
}
```

And use this extension to close the repl here

src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/impl/ReplForJupyterImpl.kt
```diff
@@ -639,4 +640,8 @@ class ReplForJupyterImpl(
 
     override val host: KotlinKernelHost?
         get() = currentKernelHost
+
+    override fun close() {
+        notebook.close()
```
ileasile, line 645: The same thing I've suggested for MessageHandlerImpl

ileasile (changes requested):

jupyter-lib/shared-compiler/src/main/kotlin/org/jetbrains/kotlinx/jupyter/repl/notebook/impl/NotebookImpl.kt
nikolay-egorov: Agreed, let's do it

src/main/kotlin/org/jetbrains/kotlinx/jupyter/messaging/MessageHandlerImpl.kt
nikolay-egorov: That's indeed nice move, +1

PR #453, opened 2024-02-29
KTNB-284 Add support for passing in-memory values as display data or execution results

cmelchior:
This PR is the kernel part of https://youtrack.jetbrains.com/issue/KTNB-284

This PR extends the Jupyter messages being sent from the Server to the Client with a new `inMemoryResult`. This field only makes sense for the embedded kernel used in IntelliJ, as it contains the display result as a pure in-memory object rather than serializing it to JSON.

In-memory results are represented using the `InMemoryMimeTypedResult` display result. This type also contains fallback data that will be used by the front-end when the in-memory result isn't available, like when saving a file to disk. Currently, the fallback data is a screenshot of the UI.

In terms of public user facing API's. This PR also introduces a new helper method called `SWING()` which can be used in combination with the `DISPLAY` helper function.

Usages:

```
// Cell 1: 
import java.awt.Color
import java.awt.Dimension
import java.io.File
import javax.swing.JButton
import javax.swing.JFrame
import javax.swing.JPanel

val frame = JFrame("Color Changer")
frame.defaultCloseOperation = JFrame.EXIT_ON_CLOSE
frame.preferredSize = Dimension(300, 200)
val panel = JPanel()
panel.preferredSize = Dimension(300, 200)
val button1 = JButton("Button 1")
val button2 = JButton("Button 2")
val button3 = JButton("Button 3")
button1.addActionListener { panel.background = Color.RED }
button2.addActionListener { panel.background = Color.GREEN }
button3.addActionListener { panel.background = Color.BLUE }
panel.add(button1)
panel.add(button2)
panel.add(button3)
frame.add(panel)
frame.pack()
panel

// Output1: 
<panel> will be rendered

// Cell2:
SWING(panel)

// Output2: 
<panel> will be rendered, similar to output1

// Cell3:
DISPLAY(SWING(panel))

// Output3: 
<panel> will be rendered, similar to output1, but multiple DISPLAY results can be sent.
```

**TODO:**
- [ ] Review the architecture for sending events this way. 
- [ ] Review the mimetype-definitions. Should we do it differently?
- [ ] Add better tests. Currently testing as been done by-hand.
- [ ] Should we bump the kernel version?

### DIFF
--- a/jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/defaultRenderers.kt
+++ b/jupyter-lib/api/src/main/kotlin/org/jetbrains/kotlinx/jupyter/api/defaultRenderers.kt
@@ -7,17 +7,27 @@ import java.io.ByteArrayOutputStream
 import java.lang.reflect.Array
 import java.util.*
 import javax.imageio.ImageIO
+import javax.swing.JFrame
+import javax.swing.JPanel
 
-val bufferedImageRenderer = createRenderer<BufferedImage> {
+/**
+ * Convert a buffered image to a PNG file encoded as a Base64 Json string.
+ */
+fun encodeBufferedImage(image: BufferedImage): JsonPrimitive {
     val format = "png"
     val stream = ByteArrayOutputStream()
-    ImageIO.write(it, format, stream)
+    ImageIO.write(image, format, stream)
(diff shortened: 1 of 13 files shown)


cmelchior:
Moving review to Space

Issue #470, opened 2024-10-31
Update maven-dependecy resolver to fix authentication through maven mirrors


exaV:
Could you please upgrade [kotlinMavenResolver](https://github.com/Kotlin/kotlin-jupyter/blob/e607cf16c49d25507b9a983089b507dae46706f4/gradle/libs.versions.toml#L7C1-L7C20) from 2.0.0-Beta5 to 2.0.20?  
I submitted a fix for dependency resolution with (authenticated) maven mirrors here https://github.com/JetBrains/kotlin/commit/9c05493cb10be14305a409fa8d78810c9aea1612 which is present in 2.0.20. With the fix we would finally be able to use Kotlin notebooks in our corporate environment where we (must) use maven mirror for all requests.

The two commits from the fix are also the only changes between [2.0.0-Beta5](https://github.com/JetBrains/kotlin/commits/2.0.0-Beta5/libraries/scripting/dependencies-maven) and [2.0.20](https://github.com/JetBrains/kotlin/commits/2.0.20/libraries/scripting/dependencies-maven), so it should be safe to upgrade.

cmelchior:
This should be fixed by https://github.com/Kotlin/kotlin-jupyter/commit/d91d0a1b9151653a4ee2f9d8504a36a84644fb14

exaV:
@cmelchior  First of all thank you for taking the time to apply the fix. I just tried to test the latest release but I don't know how. When I create a kotlin-notebook in intellij it does not use the latest release of kotlin-jupyter right? Can I somehow tell to IDE to do that? Or do I have to wait for a plugin update?

Another question, I know it only loosely related, but maybe you can help point me in the right direction. I have the same issue with regular *.main.kts (kotlin script) files inside Intellij because intellij uses an embedded version of the compiler (version 1.9.x) to run script files. I have no idea where I can open an issue to ask for an upgrade or track progress on that work. I my PR I asked the same and was told it would arrive with Intellj 2024.2.x but it didn't and latest EAP does not have it either.

exaV:
Thanks, that worked 👍 

[...]

PR #471, opened 2024-10-31
Upgrade kotlinMavenResolver to 2.0.20

exaV:
see https://github.com/Kotlin/kotlin-jupyter/issues/470


cmelchior:
Sorry for the late reply. I noticed that 2.0.21 was the newest version. So I ran tests on that and they checked out. So I ended up updating to that version instead. It has been merged to `master` and should be available tomorrow. The build should show up here when available: https://github.com/Kotlin/kotlin-jupyter/tags

